PR #70, opened 2021-04-01
OAC mapper

amywieliczka:


barbarahui (approved):
Looks good! I need to flesh out the nuxeo mapper to map some of the more complicated fields and I think it makes sense to use the code patterns that you have here. So I think it'll make sense to move more generalizable code into a main `Fetcher` class, i.e. create a fetcher.py file ?

PR #74, opened 2021-04-05
Oac fetcher

barbarahui:


amywieliczka (approved):
Looks good to me! Curious about @tingletech's feedback on the OAC fetcher URL - he mentioned there's maybe a different OAC URL to use? Looks like this is the one currently used in the collection registry: http://dsc.cdlib.org/search?facet=type-tab&style=cui&raw=1&relation=ark:/13030/tf1z09n955

barbarahui:
> Looks good to me! Curious about @tingletech's feedback on the OAC fetcher URL - he mentioned there's maybe a different OAC URL to use? Looks like this is the one currently used in the collection registry: http://dsc.cdlib.org/search?facet=type-tab&style=cui&raw=1&relation=ark:/13030/tf1z09n955

Hmm yeah I did talk to Brian about that briefly a while back, and I think he said that he vaguely remembers deciding with Mark that we should use that fetcher URL -- but he doesn't remember exactly what the logic was anymore, so we might as well leave it :D Shall we ask him again?

PR #107, opened 2021-06-25
Move s3 bucket and key names into env vars

barbarahui:


barbarahui:
This was a pretty straightforward code change so I'm merging it myself...

PR #139, opened 2022-01-29
Complex fetcher

barbarahui:
Essentially re-creating https://github.com/ucldc/rikolti/pull/129 after merging in https://github.com/ucldc/rikolti/pull/133 in error and having to back things out.


barbarahui:
I am going ahead and merging this in without Amy reviewing since it's a copy of https://github.com/ucldc/rikolti/pull/129

PR #185, opened 2022-10-18
Fix import statements

barbarahui:
Hey @amywieliczka -- I think this is the correct fix for the import statements? 

Also, I figure using github like this is the easiest way to communicate about these kinds of things, but lmk if a different workflow seems better :)

### DIFF
--- a/metadata-mapper/validate_mapping.py
+++ b/metadata-mapper/validate_mapping.py
@@ -3,8 +3,8 @@ import os
 import boto3
 import requests
 import json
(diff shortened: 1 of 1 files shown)


amywieliczka (approved):
Thanks for catching this @barbarahui!

I definitely made changes to the mapper without making changes to this file since I wasn't sure if we would use validate_mapping. It's probably best to either keep validate_mapping up-to-date, or remove it from the repository, though. Thanks for keeping it up to date.

PR #186, opened 2022-10-18
lxml library question

barbarahui:
Hey @amywieliczka , would it be OK to remove the lxml library from the repo and install it via pip instead? Of course now I'm looking at your commit message (https://github.com/ucldc/rikolti/commit/7ca87d3ba821324f27a15c986288b129b26f539c), I'm surmising that you needed to do this to get the mapper to work in lambda?

The problem is that sickle, which I'm using for the OAI mapping, tries to import lxml and I get the following error:

(.env) bhui@cdl-bhui-9m metadata-mapper % python tests.py
Traceback (most recent call last):
  File "/usr/local/dev/rikolti/metadata-mapper/tests.py", line 1, in <module>
    from lambda_function import lambda_handler
  File "/usr/local/dev/rikolti/metadata-mapper/lambda_function.py", line 8, in <module>
    from islandora_oai_dc_mapper import IslandoraVernacular
  File "/usr/local/dev/rikolti/metadata-mapper/islandora_oai_dc_mapper.py", line 6, in <module>
    from sickle import models
  File "/usr/local/dev/rikolti/metadata-mapper/.env/lib/python3.9/site-packages/sickle/__init__.py", line 13, in <module>
    from sickle.response import OAIResponse
  File "/usr/local/dev/rikolti/metadata-mapper/.env/lib/python3.9/site-packages/sickle/response.py", line 9, in <module>
    from lxml import etree
ImportError: cannot import name 'etree' from 'lxml' (/usr/local/dev/rikolti/metadata-mapper/lxml/__init__.py)

I can probably figure out a workaround, but just wanted to check before going down that road.


amywieliczka:
We do need to keep lxml in the repo (or write a complex and expensive deployment script that involves spinning up an ec2 instance every single time). 

I am a bit confused why your virtual environment is trying to import lxml from `/rikolti/metadata-mapper/lxml/` but I'm not sure of the specifics of how your machine is set up. While we do need to keep this lxml in the repository, and it is the version of lxml that we need to deploy to AWS Lambda, you should be running a different lxml (built against your machine, not an amazon linux 2 machine) in your local environment.

barbarahui:
@amywieliczka When python looks for modules, it looks first in the current directory, so since lxml is in the current directory (`/rikolti/metadata-mapper/lxml/`), it tries import that. I do have lxml installed in my virtual env.

barbarahui:
@amywieliczka I think the solution is to use AWS SAM to build and deploy the code to lambda. I've taken a crack at it, and I think it will work:

https://github.com/ucldc/rikolti/tree/aws-sam-mapper/metadata-mapper

barbarahui:
Closing this pull request without merging. Addressing this issue by using AWS SAM as part of this pull request: https://github.com/ucldc/rikolti/pull/190

PR #191, opened 2022-10-22
Enrichment Chain Framework + first set of enrichments

amywieliczka:
Migrated enrichments as checked off on this wiki page: https://github.com/ucldc/rikolti/wiki/Enrichments


barbarahui (approved):
This approach makes sense to me, i.e. do a fairly straight port of the code from DPLA for now. Oof there is a lot to rationalize at some point though. It also makes sense to me in terms of how you've incorporated the enrichments into the mapper code here. I feel like I don't have super substantive comments at this point otherwise. Really the main priority is getting the code ported over right now, and it looks like this is going fine.

amywieliczka:
Apply a function name prefix to enrichments called from enrichment chain and another function name prefix to enrichment helper functions called from within an enrichment. `dpla__` and `dpla_helper__`

PR #230, opened 2022-11-30
Sam rationalize 2

barbarahui:


README.md
```diff
@@ -66,3 +66,27 @@ We use PR reviews to approve or reject, comment on, and request further iteratio
...
+
+```
+sam build
+```
+
+> **NOTES:**
+>
+> There is a [troposphere](https://troposphere.readthedocs.io/en/latest/quick_start.html) script stub named `create_sam_template.py` checked into the repo. This script generates `template.yaml`, but we decided not to use it for now since the template is simple and we don't need to introduce another layer of tooling at this point. We might use it in the future to generate templates for different environments and such.
+
+> We will eventually want to use `sam build` with the `--use-container` option for compiling the `lxml` dependency for the `metadata_mapper`, as lxml needs to be natively compiled, i.e. has C bindings and is not pure python.
```
amywieliczka, line 86: Is there a blocker on figuring this out? Shouldn't we just figure it out now while we're in the midst of all this SAM stuff? Sorry - I guess what I meant this morning was we didn't need to do it together on the call, but we should probably get this sorted now? Unless you just want to make a new card to add to your stack for next sprint?

amywieliczka (approved):
Let's either make a card for looking into the --use-container flag or do that now as part of this PR - up to you. Besides that small bit that is incomplete, this PR looks good though and approving for merging.

README.md
barbarahui: Oh yeah I totally misunderstood what we were saying yesterday! I will figure this out and add it to the README before we merge this in (:

barbarahui: @amywieliczka OK it looks like it's pretty straightforward to use `sam build --use-container`. I updated the README with instructions for doing so.

amywieliczka (approved):

README.md
amywieliczka: wow, you don't even have to provide a container. that's some magic

PR #236, opened 2022-12-02
Mapper Validation

amywieliczka:
Validation is still in draft format. Major updates:

- Previously, fetcher.increment() returned None if we couldn't increment pages any further, and returned a json representation of the fetcher state if we could. Now, fetcher.increment() returns the json representation of the fetcher state regardless, but when we can't increment pages any further, it adds the key 'finished': True to that fetcher state dictionary. 
- Added python scripts fetch_registry_collections.py, map_registry_collections.py, and validate_registry_collections.py - these are helper scripts that take a registry API endpoint, for example:
```bash
python fetch_registry_collections.py "https://registry.cdlib.org/api/v1/rikoltifetcher/?format=json&harvest_type=NUX"
python map_registry_collections.py "https://registry.cdlib.org/api/v1/rikoltimapper/?format=json&mapper_type=ucldc_nuxeo"
python validate_registry_collections.py "https://registry.cdlib.org/api/v1/rikoltimapper/?format=json&mapper_type=ucldc_nuxeo"
```

These API endpoints all return a list of collection records properly formatted for input as payloads to the fetchers & mappers lambdas. Careful running them and walking away! There are 322 collections at the endpoints represented above. You'll likely only want a subset of all of them. If you can express your subset as a filter on the underlying registry collection queryset using these fields: https://github.com/ucldc/avram/blob/master/library_collection/api.py#L44, then you can create a registry API endpoint for that query, and run these scripts against that registry API endpoint.


metadata_fetcher/fetchers/oai_fetcher.py
```diff
@@ -102,15 +102,16 @@ def increment(self, http_resp):
...
             "harvest_type": self.harvest_type,
             "collection_id": self.collection_id,
             "write_page": self.write_page,
             "harvest_data": self.oai
-        })
+        }
+        if not self.oai.get('resumption_token'):
+            finished = {"finished": True}
+            finished.update(current_state)
+            return json.dumps(finished)
```
barbarahui, line 115: This is a little confusing as is. How about more simply: 

```
        if not self.oai.get('resumption_token'):
            current_state.update({"finished": True})

        return json.dumps(current_state)
```

metadata_fetcher/fetchers/oac_fetcher.py
```diff
@@ -112,12 +112,15 @@ def increment(self, http_resp):
...
             "write_page": self.write_page,
             "harvest_data": self.oac
-        })
+        }
+        if not self.oac.get('current_group'):
+            finished = {"finished": True}
+            finished.update(current_state)
+            return json.dumps(finished)
+
+        return json.dumps(current_state)
```
barbarahui, line 126: Similarly, how about:
```
        if not self.oac.get('current_group'):
            current_state.update({"finished": True})

        return json.dumps(current_state)
```

metadata_fetcher/fetchers/Fetcher.py
```diff
@@ -62,8 +71,12 @@ def fetchtos3(self, page):
...
     def fetch_page(self):
         page = self.build_fetch_request()
-        response = requests.get(**page)
-        response.raise_for_status()
+        try:
+            response = requests.get(**page)
+            response.raise_for_status()
+        except requests.exceptions.HTTPError:
+            raise FetchError(
+                f"[{self.collection_id}]: unable to fetch page {page}")
```
barbarahui, line 79: Would it be helpful to include the error message from requests in the output to give some insight into what went wrong?

barbarahui (changes requested):
Looks good. Just a few minor comments/questions.

metadata_fetcher/fetchers/oai_fetcher.py
amywieliczka: oh. yeah. totally! thanks @barbarahui

barbarahui (approved):

PR #243, opened 2022-12-15
Tf/cca vault mapper

lthurston:
This PR is dependent on https://github.com/ucldc/rikolti/pull/282


metadata_mapper/lambda_function.py
```diff
@@ -17,16 +18,19 @@ def import_vernacular_reader(mapper_type):
...
 
-    mapper_type_words = mapper_type.split('_')
+    mapper_type_words = snake_mapper_name.split('_')
     class_type = ''.join([word.capitalize() for word in mapper_type_words])
     vernacular_class = getattr(
         mapper_module, f"{class_type}Vernacular")
 
-    if vernacular_class not in VernacularReader.__subclasses__():
-        print(f"{ mapper_type } not a subclass of VernacularReader")
+    if not issubclass(vernacular_class, AbstractVernacular):
```
amywieliczka, line 32: A simple issubclass won't quite work here (IslandoraVernacular is a subclass of OaiVernacular is a subclass of AbstractVernacular), but I've added a recursive implementation `all_subclasses` here: https://github.com/ucldc/rikolti/pull/244/files (along with heaps of bug fixes in the enrichments)

timfrazee: `issubclass` looks down the entire ancestor chain for direct, indirect, or virtual subclasses. To the best of my understanding, `__subclasses__()` only looks at direct descendents, which is why I made the change since our mapper implementations (such as `CcaVaultVernacular`) are now separated by two levels.

metadata_mapper/lambda_function.py
```diff
@@ -3,7 +3,7 @@
 from urllib.parse import urlparse, parse_qs
 import settings
 import importlib
-from mappers.abstract_mapper import UCLDCWriter, AbstractRecord, AbstractVernacular
+from mappers.mapper import UCLDCWriter, Record, Vernacular
```
amywieliczka, line 6: Thanks @timfrazee - we've decided module name `mapper` and class names `Record` and `Vernacular` are what we would like to use (as you have here).

PR #244, opened 2022-12-15
Mapper & Enrichment Bug Squashing

amywieliczka:


metadata_mapper/lambda_function.py
```diff
@@ -56,40 +75,37 @@ def map_page(payload, context):
...
+            if report in group_by_report:
+                group_by_report[report].append(couch_id)
+            else:
+                group_by_report[report] = [couch_id]
+        count_by_report = {
+            report: f"{len(couch_ids)} of {len(mapped_records)}"
+            for report, couch_ids in group_by_report.items()
+        }
+        for report, count in count_by_report.items():
+            print(f"{count} records report enrichments errors: {report}")
```
amywieliczka, line 107: I've removed print statements that reported errors in each enrichment function and instead appended those errors to an array `enrichment_report` that exists on each record. This bit of logic here groups like lists, so if, for example, every single item on the page we're currently mapping has the exact same enrichment error, it's a little clearer to see. 

Rather than printing, this could get sent to the return value of map_page. We should maybe try to rationalize where to use printing vs. log messages vs. return values. I was trying to limit printing to command line utilities such as `map_registry_collections.py` and cases when execution should halt, rather than reporting something out and continuing to process. I was trying to only use log messages or return values in the internals such as this. I'm punting a bit on doing this rationalization until we have a clearer sense of our workflow management tooling.

metadata_mapper/lambda_shepherd.py
```diff
@@ -88,6 +96,16 @@ def map_collection(payload, context):
                 InvocationType="Event",  # invoke asynchronously
                 Payload=json.dumps(payload).encode('utf-8')
             )
+    return {
+        'statusCode': 200,
+        'body': {
+            'collection_id': collection_id,
```
amywieliczka, line 102: If we did send those enrichment exceptions in the return value, they could trickle up to get added to this return value.

metadata_mapper/map_registry_collections.py
```diff
@@ -8,15 +8,26 @@
 
 def map_endpoint(url):
     lookup = {
-        'ucldc_nuxeo': 'nuxeo'
+        'ucldc_nuxeo': 'nuxeo',
+        'oac_dc': 'oac',
+        'islandora_oai_dc': 'islandora'
```
amywieliczka, line 13: this sort of translation should really happen on the registry side of things, and at some point when we have a firmer "mapper vocabulary" (probably after Rikolti's launch) we should migrate the registry to use the new vocabulary.

metadata_mapper/mappers/islandora_mapper.py
```diff
@@ -58,6 +58,7 @@ def to_UCLDC(self):
         # print(f"{self.source_metadata=}")
 
         self.mapped_data = {
+            'calisphere-id': self.legacy_couch_db_id.split('--')[1],
```
amywieliczka, line 61: I'm not exactly sure what the calisphere id should be, but validation requires one. Ideally it should probably match what is currently used for the calisphere.org/item/<item_id> url, but I still don't understand where that comes from when it's not a UUID or an ARK. We should either choose to take this on ourselves as a mapper update we'll have to make at some point across the entire suite of mappers (and remove it from validation requirements), or discuss with CIC.

metadata_mapper/mappers/mapper.py
```diff
@@ -66,6 +66,8 @@ class Record(object):
     def __init__(self, col_id, record):
         self.collection_id = col_id
         self.source_metadata = record
+        self.pre_mapped_data = {}
+        self.enrichment_report = []
```
amywieliczka, line 70: This is that enrichment exceptions error reporting thing. 

pre_mapped_data exists for the sake of some enrichments that happen prior to mapping. it's a stop gap. we should get rid of it.

metadata_mapper/mappers/mapper.py
```diff
@@ -137,9 +140,12 @@ def select_oac_id(self):
...
             calisphere_id = id_values[0]
 
-        self.mapped_data["id"] = f"{self.collection_id}--{calisphere_id}"
-        self.mapped_data["isShownAt"] = calisphere_id
-        self.mapped_data["isShownBy"] = f"{calisphere_id}/thumbnail"
+        self.pre_mapped_data["id"] = f"{self.collection_id}--{calisphere_id}"
+        self.pre_mapped_data["calisphere-id"] = f"{calisphere_id}"
+        self.pre_mapped_data["isShownAt"] = calisphere_id
+        self.pre_mapped_data["isShownBy"] = f"{calisphere_id}/thumbnail"
+        self.legacy_couch_db_id = (f"{self.collection_id}--{calisphere_id}")
```
amywieliczka, line 147: Currently the validation outputs the legacy couchdb id as a way to look up how the record looks in our existing system. Calculating this legacy couchdb id in rikolti helps us match against the existing system. Unlike our solr ids and "calisphere-id", I do understand how the couchdb id is calculated each time. I'd prefer to have both in the system until we can get rid of both solr and couch.

metadata_mapper/mappers/mapper.py
```diff
@@ -219,7 +225,7 @@ def required_values_from_collection_registry(
             # self.mapped_data["@context"] = "http://dp.la/api/items/context"
         return self
 
-    def shred(self, field, delim=";"):
+    def shred(self, prop, delim=";"):
```
amywieliczka, line 228: This was just incorrect naming.

metadata_mapper/mappers/mapper.py
```diff
@@ -231,16 +237,16 @@ def shred(self, field, delim=";"):
...
+        1,021 times:    prop=["sourceResource/creator"]
+        1,021 times:    prop=["sourceResource/type"]
+        5 times:        prop=["sourceResource/description"],   delim=["<br>"]
         """
-        field = field[0].split('/')[1:]     # remove sourceResource
+        field = prop[0].split('/')[1:][0]     # remove sourceResource
         delim = delim[0]
 
-        if field not in self.mapped_data:
+        if field not in self.mapped_data or not self.mapped_data[field]:
```
amywieliczka, line 249: sometimes field exists in self.mapped_data but self.mapped_data[field] literally equals None. Would love to understand better if there are cases where a "None" field value is actually important to know, or if this is just a side effect of implementation and we can safely clean up these fields.

metadata_mapper/mappers/mapper.py
```diff
@@ -308,8 +313,8 @@ def copy_prop(self, prop, to_prop, skip_if_exists=[False]):
                     to_prop=["sourceResource/title"]
         """
 
-        src = prop[0].split('/')[1:]
-        dest = to_prop[0].split('/')[1:]
+        src = prop[0].split('/')[-1]
+        dest = to_prop[0].split('/')[-1]
```
amywieliczka, line 317: this update handles `to_prop = ["sourceResource/title"]` AND `to_prop = ["dataProvider"]`

(though obviously still unclear on what data provider means in the to_prop category, we'll have to wait till we harvest a collection that uses this enrichment with these parameters)

metadata_mapper/mappers/mapper.py
```diff
@@ -351,18 +355,22 @@ def move_date_values(self, prop, dest="temporal"):
         # 2079 times: prop=["sourceResource/spatial"]
         """
         src = prop[0].split('/')[-1]   # remove sourceResource
-        if src not in self.mapped_data:
+        src_values = self.mapped_data.get(src)
```
amywieliczka, line 358: this is that same issue where sometimes src is in self.mapped_data but it is actually a None value.

metadata_mapper/mappers/mapper.py
```diff
@@ -351,18 +355,22 @@ def move_date_values(self, prop, dest="temporal"):
...
             return self
-
-        src_values = self.mapped_data[src]
         if isinstance(src_values, str):
             src_values = [src_values]
-        remove = []
+
         dest_values = self.mapped_data.get(dest, [])
+        if not dest_values:
+            dest_values = []
```
amywieliczka, line 366: same issue with keys for None values

metadata_mapper/tests.py
```diff
@@ -1,19 +1,59 @@
...
-for harvest in harvests:
-    print(f"mapping tests: {json.dumps(harvest)}")
-    map_page(json.dumps(harvest), {})
-    print(f"mapped: {str(harvest)}")
+def main():
+    vernacular_path = settings.local_path(
+        'vernacular_metadata', 0)[:-1]
+    urls = [
+        f"https://registry.cdlib.org/api/v1/rikoltimapper/{f}/?format=json"
+        for f in os.listdir(vernacular_path)
```
amywieliczka, line 21: this now just maps against everything in the local filesystem's vernacular metadata path.

metadata_mapper/validate_mapping.py
```diff
@@ -207,6 +205,22 @@ def validate_mapped_page(rikolti_records, solr_records, query):
                     )
 
         query['start'] = query['start'] + 100
+        if query['start'] < num_solr_records:
+            collections_search = solr(**query)
+            collections_search = collections_search.get(
+                'response', {'docs': None}).get('docs', [])
+            solr_records.update(
+                {r['harvest_id_s']: r for r in collections_search})
+        else:
```
amywieliczka, line 214: Add handling for cases when Rikolti has new records that don't appear in Solr

metadata_mapper/lambda_function.py
barbarahui: Sounds good. I have wondered about when to use print statements vs logging vs some other reporting method also. Figuring out a clear rationalization at some point sounds great.

metadata_mapper/map_registry_collections.py
barbarahui: makes sense

metadata_mapper/lambda_shepherd.py
barbarahui: Yeah that could be useful

metadata_mapper/lambda_shepherd.py
```diff
@@ -55,24 +56,31 @@ def map_collection(payload, context):
...
+            page_list = [f for f in os.listdir(vernacular_path)
+                         if os.path.isfile(os.path.join(vernacular_path, f))]
+        except FileNotFoundError as e:
+            logging.debug(f"{e} - have you fetched {collection_id}?")
+            return {
+                'statusCode': 400,
+                'body': json.dumps({
+                    'error': (
+                        f"{repr(e)} - have you fetched {collection_id}? ",
+                        f"looked in {e.filename}"
```
barbarahui, line 74: I was confused about this line at first, but it's reporting the name of the directory we looked for files in, yes? Super minor, but maybe this would make it clearer when reading the code:

```
Looked in dir {e.filename}
```

metadata_mapper/lambda_function.py
```diff
@@ -45,6 +45,25 @@ def parse_enrichment_url(enrichment_url):
     return enrichment_func, kwargs
 
 
+def run_enrichments(records, payload, enrichment_set):
+    collection = payload.get('collection', {})
+    for enrichment_url in collection.get(enrichment_set, []):
+        enrichment_func, kwargs = parse_enrichment_url(enrichment_url)
+        if not enrichment_func and settings.SKIP_UNDEFINED_ENRICHMENTS:
```
barbarahui, line 52: Is it possible for enrichment_func to be None and for settings.SKIP_UNDEFINED_ENRICHMENTS to be False? Do we want to handle that case?

metadata_mapper/tests.py
barbarahui: oh yeah that's nicer

metadata_mapper/validate_mapping.py
barbarahui: Makes sense. Wow yeah, the (necessary) logic for getting pages from solr and comparing to pages of rikolti records is convoluted...

metadata_mapper/mappers/islandora_mapper.py
barbarahui: Yeah, we need to do some tracing of code through the old stack to see where this is set. I *think* that if the item doesn't have an ARK, then we use the originating system's id (which in some cases is a UUID).

barbarahui: I think this would be a good candidate task for CiC to work on?

metadata_mapper/mappers/oac_mapper.py
```diff
@@ -9,35 +9,22 @@
...
         })
 
-        mapped_data = self.remove_if_none([
-            "provider",
-            "dataProvider",
-            "isShownBy",
-            "item_count",
-        ], mapped_data)
+        if mapped_data["isShownBy"] is None:
+            mapped_data.pop("isShownBy")
```
barbarahui, line 24: Do we need to do the same for if isShownAt is None?

metadata_mapper/mappers/mapper.py
barbarahui: got it

metadata_mapper/mappers/mapper.py
barbarahui: ok, makes sense

metadata_mapper/mappers/mapper.py
barbarahui: Apparently Lucene (which is what solr and elasticsearch are both built on) cannot by definition index a null value. It could index the string "None", but I'm doubtful that this is ever what is intended? We could check with Adrian?

barbarahui (approved):
Wow, thanks again for doing this huge amount of work to migrate the enrichments. They are so...gnarly. I made a few comments, lots of which are questions.

metadata_mapper/mappers/islandora_mapper.py
amywieliczka: I've never seen the calisphere/solr id stored in couchdb. You can get the couchdb id from solr, but not the other way around. I think the solr id must currently be created on the sync to solr step (so I don't really think this is within scope for CiC?)

metadata_mapper/lambda_function.py
amywieliczka: In `parse_enrichment_url` I raise an exception if SKIP_UNDEFINED_ENRICHMENTS is False and enrichment_func is None.

metadata_mapper/mappers/oac_mapper.py
amywieliczka: OAC mapper kind of needs to be completely revisited. At some point. I don't think this makes sense to look at for now though.

metadata_mapper/lambda_shepherd.py
amywieliczka: Done

barbarahui (approved):

PR #286, opened 2023-01-19
Implement cca_vault mapper

lthurston:
Draft because I need to test it, and squash.


lthurston:
Merging into mapper restructure for simplicity.

PR #288, opened 2023-01-20
Exclude map param when not present

lthurston:
The &set=None parameter was preventing fetch
from interviews.televisionacademy.com


lthurston:
@barbarahui here's that little fix we discussed in Slack

amywieliczka (approved):
Thanks Lucas!

PR #298, opened 2023-01-23
Deprecate mapper methods

lthurston:
I meant to include this in the mapper refactor PR but never pushed it. Not urgent, can follow other things.


lthurston:
@amywieliczka @barbarahui This is a fairly trivial change to properly identify deprecated methods. It makes the output a little noisier, which will force us to address it when we get to mappers that need to be changed.

amywieliczka (approved):

PR #299, opened 2023-01-23
Mapper helper funcs

lthurston:
Moving some mapper helper methods into functions. This PR follows #298 which is why it has the commit from that one too.  Linting was applied too.

### DIFF
--- a/metadata_mapper/mappers/mapper.py
+++ b/metadata_mapper/mappers/mapper.py
@@ -69,6 +69,28 @@ class Vernacular(ABC, object):
         return api_response
 
 
+# Static mapping helpers
+def collate_plucked_values(values: list, pluck: str) -> list:
+    return [f[pluck] for f in values]
+
+
+def collate_values(values):
+    collated = []
+    for value in values:
+        if not value:
+            continue
+
+        if isinstance(value, str):
+            collated.append(value)
+        else:
+            collated.extend(value)
+    return collated
+
+
+def first(item):
+    return item[0] if item else None
+
+
 class Record(ABC, object):
 
     def __init__(self, collection_id: int, record: dict[str, Any]):
@@ -128,20 +150,6 @@ class Record(ABC, object):
         pass
 
     # Mapper Helpers
-    def collate_plucked_values(self, values: list, pluck: str) -> list:
-        return [f[pluck] for f in values]
-
-    def collate_values(self, values):
-        collated = []
-        for value in values:
-            if not value:
-                continue
-
-            if isinstance(value, str):
-                collated.append(value)
-            else:
-                collated.extend(value)
-        return collated
 
     @deprecated(reason="replace with `collate_plucked_values()` in mappers that use it")
     def collate_subfield(self, field, subfield):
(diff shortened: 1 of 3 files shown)


lthurston:
This has been rebased on top of main, merge conflicts addressed.

amywieliczka (changes requested):
IMO, the Record base class _should_ be a collection of common mapping functions that might occur as operations on a Record. Currently it is a bit of a hodgepodge, but I would like to move in that direction. Moving collate_values outside of the Record base class is actually a move away from that direction. 

Though certainly, I don't know what "Record.collate_values()" means. Records have fields which have values. It makes semantic sense to say "Record.collate_fields", but "Record.collate_values" makes a lot less sense. I don't know exactly how to resolve this. What is your vision here?

lthurston:
Thanks for your feedback, @amywieliczka . As I've mentioned, I'm pretty green with Python, so I'm fully aware that my opinions on this are almost certainly based on experience with paradigms in other languages and may not be the most Pythonic. With that in mind, I'll explain my thinking.

In a [previous, merged commit](https://github.com/ucldc/rikolti/commit/19d8901c7007499fbbaf62996eb55445fb379244), we teased apart the two things that the original `collate_fields` function was doing (1. getting values from `source_metadata` and 2. building a dictionary of the "truthy" ones), this in an effort to make the mapping code more explicit and flexible. Once we have the values we need, the building of the dictionary itself is static and doesn't depend on the state of the `Record` instance at all, so my instinct was to make these methods (referring specifically to `collate_values`, `collate_plucked_values`) explicitly static methods in the class. That's what I would have done in another language, but in Python, with the module as a primary organizational and namespacing unit, the static method approach still felt unnecessarily associated with the `Record` class despite not having any need for any of the context that would be passed in as the first argument either as an instance method or a class method. So, I extracted these methods from the class, leaving them still explicitly associated with mapping as a pure function in the module, but not bound to the Record class or an instance of it.

My goal isn't to make it a hodgepodge, of course, but I can certainly see that perspective. My goal is to keep the code organized, but also have it exist in the appropriate scope based on what it does, and what its dependencies are.

To answer your second question, I agree that `collate_values` doesn't make a ton of sense to me either. Looking back at [this conversation](https://github.com/castiron/rikolti/pull/3), I'm seeing that you agreed to renaming these `compose_*`. I missed this, and apologize. If I `s/collate/compose/g`'d these function names, are we back on track? Again, sorry for missing that response.

lthurston:
Quick follow up: going with the `@staticmethod` approach would be fine too, but more verbose than if these were just functions. As I work on the ContentDM mapper, there are a handful of additional mapper helper methods that I've written, for now leaving them as instance methods while we work this out:

```
    def get_first_match(self, items: list, match: str):
        """Given a list and a match string, returns the first match
        """
        return first(self.get_matches(items, match))

    def get_matches(self, items: list, match: str):
        """Given a list and a match string, returns those dictionary items containing the string
        """
        [item for item in items if match in item]

    def flatten_list(self, items: list):
        """Given a list, flattens and strips the list items
        """
        return list([s.strip() for s in itertools.chain.from_iterable(items)])

    def split_and_flatten(self, items: list, split: str):
        """Given a list of strings or nested lists, splits the values on the split string then flattens
        """
        return self.flatten_list([c.split(split) for c in filter(None, items)])

    def ensure_list(self, item: Union[str, list]):
        """Given a list or string, ensures we have a list
        """
        return [item] if isinstance(item, str) else item
```

Calling these in a series can be noisy:

```
    def map_subject(self):
        if "subject" not in self.source_metadata:
            return

        values = self.split_and_flatten(self.ensure_list(self.source_metadata.get("subject")), self.__split_string())
        return [{'name': v} for v in values]
```

I prefer:

```
values = split_and_flatten(ensure_list(self.source_metadata.get("subject")), self.__split_string())
```

because it's clear what requires that `self` context, and what doesn't, therefore easier to quickly grok what's going on. Anyway, it's subjective, of course, and your call ultimately.

PR #300, opened 2023-01-24
Redo: Implement up_oai_dc mapper

lthurston:


lthurston:
I noticed on this mapper that the `rights` field is populated and contains HTML markup. This appears to be overwritten in Solr probably with the same enrichment that's providing that content for other mappers. However, in so doing, the HTML tags are lost, ending up with plain content like this: 

> To view additional information on copyright and related rights of this item, such as to purchase copies of images and/or obtain permission to publish them, click here to view the Holt-Atherton Special Collections policies

PR #302, opened 2023-01-25
Mapper validator

timfrazee:
Adds validator class and scaffold to support its use.

## Validator

The Validator class (`metadata_mapper.mappers.validator`) contains an `errors` property (an instance of `ValidationErrors`) that is populated when `Validator#validate` is invoked with two sets of data to compare. Once `validate` is called at least once, you can output errors in various ways.

For example:

```python
from mappers.validator import Validator

# Get a validator instance
validator = Validator()
# Run comparison on two datasets
validator.validate(unique_key, rikolti_mapped_data, comparison_data) 

# Returns an instance of ValidationErrors
validation_errors = validator.errors 
# Print errors to console
validation_errors.output("stdout")
# Output a list of lists of strings suitable for generating a CSV
validation_errors.csv_content()
# Write error CSV to bucket at path "validation/{collection_id}/{timestamp}.csv"
validation_errors.write_csv(collection_id=1234)
# Write CSV to specific file path
with open("./errors.csv", "w") as file:
  validation_errors.output_csv_to_file(file)
```

In order to easily validate a collection (or a page in a collection), convenience functions are available in `metadata_mapper/validate_mapping.py`.

```python
import validate_mapping

# Validate all pages in a collection; returns an instance of Validator or a subclass of Validator (see below)
validate_mapping.validate_collection(collection_id)

# Validate a single page in a collection
# In this case, you need to pass in an instance of a Validator
from mappers.validator import Validator
validator = Validator()
validate_mapping.validate_page(collection_id, page_id, validator)
```

`validate_mapping` can be invoked from the shell in order to automatically generate a CSV in the env-appropriate bucket:

```bash
# Runs validations for collection 1234, outputting a CSV at rikolti_bucket/validation/1234/{timestamp}.csv
python validate_mapping.py 1234
```

## Subclassing

`Validator` is designed to be subclassed in order to be able to respond to the validation needs of any given mapper.

By default, all instances of `Record` use the base `Validator` class. `Validator` is functional in and of itself, and does not need to be subclassed to work. In many cases, `Validator` will probably be sufficient.

In order to assign a validator subclass to a Record, simply set it as a class property:

```python
# mappers/some_sort_of_mapper.py

from mappers.mapper import Record
from mappers.validator import Validator

class SomeSortOfValidator(Validator):
  # Whatever overrides you need

class SomeSortOfRecord(Record):
  validator = SomeSortOfValidator
```

When running validations from `validate_mapping#validate_collection`, the script will attempt to find the appropriate validator class (if not explicitly provided as `validator_class`) from the mapper module name.

Most likely, if you're subclassing `Validator`, you'll want to override the methods `#set_validatable_fields` or `generate_keys`.

### set_validatable_fields(self, fields: list[dict] = [], merge: bool = True)
This method is called in `__init__` and sets the fields that will be compared and validated between the two datasets provided to `#validate`.

By default, `Validator` uses `default_validatable_fields` as defined in `mappers.validator.py`. However if you override this method, you can provide any values you'd like. It should set `self.validatable_fields` to a list of dicts that looks something like this:

```python
    {
        "field": "field_name",  # field name - should be the same in both datasets
        "type": str,  # expected type
        "validations": [Validator.content_match],  # a list of Callables that perform the validations on this field
        "level": "WARN" # The logging level to report a failure as
    },
    {
        "field": "another_field",
        "type": str,
        "validations": {Validator.content_match: "ERROR", Validator.type_match: "WARN"} # Alternatively validations can be a dict of Callable: str to define log levels for each individual validator
    },
```

### generate_keys(self, collection: list[dict], type: str = None, context: dict = {})
This method is called in order to generate unique keys ("harvest IDs") for a dataset. This is performed in `validate_mapping` in order to create an intersection of records - two records to be compared should generate the same key.

If a metadata source requires some custom logic to make sure the records align, this method can be overridden. Provided a list of dicts, it should return a dict with the unique keys as the dict keys, and the records (from the provided list) as the values.


lthurston:
@barbarahui @amywieliczka Tim will miss our Monday meeting but I wanted to just put this preliminary work in front of you to see how we're envisioning this.

metadata_mapper/utilities.py
```diff
@@ -0,0 +1,138 @@
...
+    """
+    accept underscored_module_name_prefixes
+    accept CamelCase class name prefixes split on underscores
+    for example:
+    mapper_type | mapper module       | class name
+    ------------|---------------------|------------------
+    nuxeo       | nuxeo_mapper        | NuxeoVernacular
+    content_dm  | content_dm_mapper   | ContentDmVernacular
+    """
+    mapper_parent_modules, snake_cased_mapper_name = mapper_type.split(".")
```
amywieliczka, line 23: `mapper_parent_modules` should be `*mapper_parent_modules`

metadata_mapper/utilities.py
```diff
@@ -0,0 +1,138 @@
...
+
+    mapper_module = importlib.import_module(
+        f"mappers.{mapper_parent_modules}.{snake_cased_mapper_name}_mapper",
```
amywieliczka, line 26: Should read: `f"mappers.{'.'.join(mapper_parent_modules)}.{snake_cased_mapper_name}_mapper",`

metadata_mapper/utilities.py
```diff
@@ -0,0 +1,138 @@
...
+
+    mapper_type_words = snake_cased_mapper_name.split('_')
+    class_type = ''.join([word.capitalize() for word in mapper_type_words])
+    vernacular_class = getattr(
+        mapper_module, f"{class_type}Vernacular")
+
+    if not issubclass(vernacular_class, Vernacular):
+        print(f"{mapper_type} not a subclass of Vernacular")
+        exit()
+    return vernacular_class
```
amywieliczka, line 38: How about just re-copy this function from main branch `lambda_function.py`?

metadata_mapper/utilities.py
```diff
@@ -0,0 +1,138 @@
...
+        page_path = os.sep.join([dir_path, str(file_name)])
+
+        with open(page_path, "a" if append else "w") as file:
+            file.write(content)
+    elif settings.DATA_SRC == 's3':
+        s3 = boto3.resource('s3')
+        bucket = 'rikolti'
+        key = f"{directory}/{collection_id}/{file_name}"
+        s3_obj = s3.Object(bucket, key)
+        s3_obj.put(Body=content)
```
amywieliczka, line 138: 🎉🙏💯💪 love this utilities module, can't wait to incorporate it into the rest of the metadata_mapper module (and maybe even make it a rikolti-wide utilities module). thank you!

metadata_mapper/validate_mapping.py
```diff
@@ -1,11 +1,22 @@
...
 import urllib3
+
+from datetime import datetime
+from pathlib import Path
+from typing import Type, Union
+
+import settings
+import utilities
+
+from map_registry_collections import MAPPER_MAP
```
amywieliczka, line 16: `lookup`? should really switch to using `rikolti_mapper_type` rather than `mapper_type` + `lookup` (or in this case, `mapper_type` + `MAPPER_MAP`

metadata_mapper/validate_mapping.py
```diff
@@ -18,330 +29,143 @@ def solr(**params):
...
-            f"[{collection_id}]: Validated page {page_path.split('/')[-1]} "
-            f"- {len(collection_report)} errors"
+            f"[Collection {collection_id}]: "
+            f"[{url}]"
+            f"{err}; A valid collection id is required for validation"
         )
+        return
+
+    mapper = collection_data.get("rikolti_mapper_type")
+    mapped_mapper = MAPPER_MAP.get(mapper, mapper)
```
amywieliczka, line 70: Shouldn't need this? `rikolti_mapper_type` is already in the format of `oai.islandora`, for example.

metadata_mapper/validate_mapping.py
```diff
@@ -18,330 +29,143 @@ def solr(**params):
...
+
+    if len(mapped_metadata) == 0 or len(solr_data) == 0:
+        print("No data found in "
+              f"{'mapped Rikolti data' if len(mapped_metadata) else 'Solr'}."
+              " Aborting."
+              )
+        return
 
-    return collection_report
+    intersection = set(mapped_metadata.keys()) & set(solr_data.keys())
```
amywieliczka, line 144: If `len(intersection) != len(mapped_metadata) != len(solr_data)`, then we should probably report out the set difference as well. Given that  `solr_data` is a function of mapped_metadata, it seems quite impossible that `len(solr_data) > len(mapped_metadata)`, so this is likely always a set difference of `set(mapped_metadata.keys()) - set(solr_data.keys())`.

Which also brings up the point that we do need to report out cases where a record exists in Solr that does not exist in Rikolti. Perhaps this is not part of the validator, since it is less a record validation and more a collection extent validation? Where would such a report fit into the framework you've presented here, though?

amywieliczka (changes requested):
@timfrazee please see inline comments. thanks so much for your thorough thinking and approach on this!

metadata_mapper/lambda_function.py
```diff
@@ -24,14 +28,15 @@ def import_vernacular_reader(mapper_type):
         package="metadata_mapper"
     )
 
-    mapper_type_words = snake_cased_mapper_name.split('_')
+    mapper_type_words = mapper_type.split(".")[-1].split('_')
```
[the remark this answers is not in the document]
timfrazee: Hmm, likely a merge error. I'll address it.

metadata_mapper/lambda_function.py
```diff
@@ -71,14 +76,15 @@ def run_enrichments(records, payload, enrichment_set):
...
-def map_page(payload, context):
+def map_page(payload: Union[dict, str], context: dict = {}):
     if settings.LOCAL_RUN and isinstance(payload, str):
         payload = json.loads(payload)
 
     vernacular_reader = import_vernacular_reader(payload.get('mapper_type'))
     source_vernacular = vernacular_reader(payload)
     api_resp = source_vernacular.get_api_response()
     source_metadata_records = source_vernacular.parse(api_resp)
+    collection = payload.get('collection', {})
```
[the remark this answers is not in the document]
timfrazee: Nothing anymore! It addressed a dev issue that has since been fixed and merged before I rebased on main. I'll remove it.

metadata_mapper/mappers/validator.py
```diff
@@ -0,0 +1,740 @@
...
+        Parameters:
+            format: str
+                "stdout" or "csv"
+            file: IO[str] (default: None)
+                A file name/path to write to (for CSV output)
+        """
+        if format == "stdout":
+            return print(self.errors)
+        elif format == "csv" and file:
+            return self.output_csv(file)
```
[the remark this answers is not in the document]
timfrazee: I've used `output` mainly for dev, to get errors printed to the console. It is not currently used in the `validate_mapping` workflow.

The original idea was to have a single interface to output, and to use an argument (`format`) to determine how it gets output. That way, different workflows could output however they'd like. However, I realized that this didn't really work for existing S3 output (which writes using boto3 - this output method requires a file reference), and instead used the existing logic (now in `utilities`) to write to the env-appropriate bucket.

We could remove `output`, but I think it would still be nice to have a convenience function to output to stdout. Perhaps something simply named `output_to_stdout`?

metadata_mapper/mappers/validator.py
```diff
@@ -0,0 +1,740 @@
...
+
+    def csv_content(self, include_fields: list[str] = None,
```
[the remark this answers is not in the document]
timfrazee: Yep!

metadata_mapper/mappers/validator.py
```diff
@@ -0,0 +1,740 @@
...
+
+        return ret
+
+    def csv_content_string(self, include_fields: list[str] = None,
+                           include_headers: bool = True) -> str:
+        content = self.csv_content(include_fields, include_headers)
+        content_strings = [",".join(row) for row in content]
+        return "\n".join(content_strings)
+
+    def output_csv_to_file(self, file: IO[str], append: bool = False,
```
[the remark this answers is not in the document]
timfrazee: Yeah, I believe this is no longer used. I'll remove it.

metadata_mapper/mappers/validator.py
```diff
@@ -0,0 +1,740 @@
...
+        return list(self.CSV_FIELDS.keys())
+
+    @property
+    def _default_headers(self) -> list[str]:
+        return list(self.CSV_FIELDS.values())
+
+
+class Validator:
+
+    def __init__(self, fields: dict = None):
```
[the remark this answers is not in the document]
timfrazee: You're right. It was originally a dict, but I found the keys to be an unnecessary complication.

I'll change the type hint, and there's also a bit of logic in building validatable fields that also appears to assume it's a dict. I'll fix that too.

metadata_mapper/mappers/validator.py
```diff
@@ -0,0 +1,740 @@
...
+                        lambda d, r, c: isinstance(r, list) and len(r) == 1
+                        ]
+    },
+    # Partial fidelity fields
+    # Content match required; nulls okay
+    {
+        "field": "alternative_title",
+        "type": str,
+        "validations": [Validator.content_match],
+        "level": "WARN"
```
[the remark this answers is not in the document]
timfrazee: It is supposed to be used as a default in `_normalize_validations`. I'll take a look and make sure that's working as intended.

metadata_mapper/utilities.py
timfrazee: You're right! I was experiementing with making this code a little less verbose, and it seems accidentally committed some of that work (which I ended up not intending to include in this branch). I'll fix it up.

metadata_mapper/validate_mapping.py
timfrazee: I think such a report could fit, if we wanted to still create a validation CSV and have rows for "Record Doesn't Exist" (I'm thinking a single row per record, not one for each key).

Or, we could optionally generate a separate "errors.csv" or something if this occurs, so it doesn't get mixed up in validation data.

metadata_mapper/validate_mapping.py
timfrazee: Yeah, not anymore. I'd imagine this should be (or has been?) addressed elsewhere, then we can rebase on that.

metadata_mapper/mappers/validator.py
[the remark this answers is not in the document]
timfrazee: Fixed - specifying a `level` at the validation definition will apply that level to all validation failures (unless otherwise specified on an individual validation)

metadata_mapper/validate_mapping.py
timfrazee: Instead of running an intersection, we now `union` the sets to get a list of all keys. If one of the datasets is missing, the validator will produce a single error and bail out.

timfrazee:
@amywieliczka @barbarahui Requested changes have been made. Please take another look as you're able.

metadata_mapper/lambda_function.py
```diff
@@ -24,14 +28,15 @@ def import_vernacular_reader(mapper_type):
         package="metadata_mapper"
     )
 
-    mapper_type_words = snake_cased_mapper_name.split('_')
+    mapper_type_words = snake_cased_mapper_name.split(".")[-1].split('_')
```
amywieliczka, line 31: Pretty sure the `.split(".")[-1]` is unnecessary here - it's already done on line 24.

metadata_mapper/mappers/validator.py
```diff
@@ -218,15 +218,27 @@ def set_validatable_fields(self, fields: list[dict] = [],
...
                 Should `fields` be merged into `default_validatable_fields`?
 
         Returns list[dict]
         """
         if merge:
-            self.validatable_fields = [*default_validatable_fields, *fields]
+            self.validatable_fields = default_validatable_fields
+
+            for field in fields:
+                index = [
```
amywieliczka, line 242: what is happening here?

amywieliczka: I read this as "index is an array of tuples, (i, el), for i el in enumerate default_validatable_fields if field['field'] == el['field']"

If index is an array of tuples, how do you use it as an index on line 239?

amywieliczka (changes requested):
Hey Tim, there's still a bit of a merge error in `lambda_function.py`, please copy the `import_vernacular_reader` function from the main branch into this file. 

Also, @barbarahui and I were both a bit confused about your merge of validatable fields in `validator.py`

metadata_mapper/lambda_function.py
timfrazee: You're right, copied over from main

metadata_mapper/mappers/validator.py
timfrazee: I agree this was too convoluted. I've simplified it a bit. Now, it performs a simple dict comprehension so it can merge (ensuring unique field names), then converts back to a list.

Refactoring validatable_fields to be a dict is an option, too, but keeping it as a list of dicts makes iterating over it a bit simpler.

amywieliczka (approved):

PR #312, opened 2023-02-03
Revert the deprecation of some methods

lthurston:
Chapman and CCA vault need to be tested / compared with Solr again before this is ready for review.

### DIFF
--- a/metadata_mapper/mappers/mapper.py
+++ b/metadata_mapper/mappers/mapper.py
@@ -1385,9 +1385,10 @@ class Record(ABC, object):
             if not solr_id:
                 solr_id = ucla_ark(couch_doc)
             if not solr_id:
-                # no recognized special id, just has couchdb id
+                if not couch_doc.get('calisphere-id'):
+                    raise Exception('no calisphere id')
                 hash_id = hashlib.md5()
-                hash_id.update(couch_doc['_id'])
+                hash_id.update(couch_doc['calisphere-id'].encode('utf-8'))
                 solr_id = hash_id.hexdigest()
             return solr_id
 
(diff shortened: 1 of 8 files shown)


amywieliczka:
You can use `calisphere-id` as the field, instead of `identifier`, and instead of `_id`, so:

```hash_id.update(str(couch_doc.get('calisphere-id', 'ABC123')).encode(encoding="UTF-8"))```

This should probably actually error out noisily if there is no calisphere-id, but we can sort that later. calisphere-id should also never be an array, so you probably don't need the `str()` or `.encode()` trappings. 

```hash_id.update(couch_doc.get('calisphere-id', 'ABC123'))```

**Comment by lthurston** [comment]
@amywieliczka Ok, all changes made here.

File: metadata_mapper/mappers/mapper.py [hunk 1]
```diff
@@ -350,7 +323,7 @@ def shred(self, prop, delim=";"):
 
         return self
 
-    def copy_prop(self, prop, to_prop, skip_if_exists=[False]):
+    def copy_prop(self, prop, to_prop, no_overwrite=None, skip_if_exists=[False]):
```
amywieliczka, line 326: did you find the one case of `no_overwrite` as a parameter??? which collection was that?

amywieliczka (approved):

metadata_mapper/mappers/mapper.py
lthurston: I didn't make a note of it, a decision I now regret. 😞

lthurston: It's here: https://registry.cdlib.org/api/v1/rikolticollection/3433/?format=json

PR #315, opened 2023-02-14
Fix enrichment issues

lthurston:


lthurston:
@amywieliczka @barbarahui These changes address some issues I encountered while testing mappers.

amywieliczka:
Thanks Lucas! This all looks great except for the lookup guard clauses - could you tell me which collection is giving you this problem? Or better yet, could you create a GitHub issue with a link to the collection throwing errors here and specify in the issue that it blocks your work on a particular mapper?

amywieliczka:
Maybe as a way to move forward, we can accept this PR as long as there is an outstanding issue with a link to the "Add guard clauses" commit. That way you're unblocked in building your mappers, but I know that ultimately this bit of code needs to be re-assessed and redone.

lthurston:
I will add an issue and reference that commit. It was an arck collection, likely the first one: https://registry.cdlib.org/api/v1/rikoltifetcher/27564/?format=json.

amywieliczka (approved):

PR #335, opened 2023-02-24
Content Harvest

amywieliczka:
Content Harvest Component

- runs in Docker, check README for run instructions

- `content_harvester/by_registry_endpoint.py` runs content harvester for registry endpoints with function `harvest_endpoint(url)`
- `content_harvester/by_collection.py` runs content harvester for a given collection with function `harvest_collection({"collection_id": 12345, "rikolti_mapper_type": "mapper_name"})`
- `content_harvester/by_page.py` runs content harvester for a given page with function `harvest_page_content(collection_id=12345, page_filename="1.jsonl" rikolti_mapper_type="mapper_name")`

`harvest_page_content`: 
1) creates a `ContentHarvester` with a persistent s3 client and http client
2) uses `get_mapped_records(collection_id, page_filename, s3_client)` to read a mapped metadata file (either locally, or on s3) and return a list of records
3) uses `ContentHarvester.harvest(record)` to harvest content for each record. 
4) warns about cases where the record has no thumbnail
5) adds a `content` key to record, value is a dictionary with all optional keys `'thumbnail', 'media', and 'children'`
6) writes the list of mapped records (either locally, or to s3) to jsonl file
7) returns a report of thumbnail source counts by mimetype and thumbnail counts by mimetype (to see how many derivatives were generated), media source counts by mimetype and media counts by mimetype (to see how many derivatives were generated), a count of children encountered while processing, and a count of the total number of records

`ContentHarvester.harvest(record)`:
1) finds the media source in the record, downloads the source to the docker container's local filesystem, and if the media source's `nuxeo_type == SampleCustomPicture`, generates a jp2 using the `derivatives` module, before optionally uploading to s3 (if `settings.CONTENT_DEST` is not `'local'`)
2) finds the thumbnail source in the record, downloads the source to the docker container's local filesystem (if it was not already downloaded by the media harvest process), and uses the `derivatives` module to make a thumbnail, before optionally uploading to s3 (if `settings.CONTENT_DEST` is not `'local'`)
3) searches for a `children` folder in the `settings.METADATA_SRC` location (locally, or on s3)
4) runs `ContentHarvester.harvest(child_record)` recursively for each child record found. 

Derivatives Module defines:
- `make_thumbnail(source_file_path, mimetype)`
- `make jp2(source_file_path, mimetype)`

Along with several helper functions.


metadata_mapper/mappers/nuxeo/nuxeo_mapper.py
```diff
@@ -165,9 +167,84 @@ def map_spatial(self):
...
+                if tv['content']['mime-type'] == 'video/mp4':
+                    file_content = tv['content']
+                    break
+
+        # this is the mapping part where we map to our data model
+        media_source = None
+        if file_content:
+            url = file_content.get('data', '').strip()
+            media_source = {
+                'url': url.replace('/nuxeo/', '/Nuxeo/'),
```
barbarahui, line 205: This line can be deleted.

content_harvester/by_page.py
```diff
@@ -0,0 +1,397 @@
...
+            Body=json.dumps(record)
+        )
+        print(f"Upload status: {upload_status}")
+
+
+def get_child_records(collection_id, parent_id, s3_client) -> list:
+    mapped_child_records = []
+    if settings.METADATA_SRC == 'local':
+        local_path = settings.local_path('mapped_metadata', collection_id)
+        children_path = os.sep.join([local_path, 'children'])
```
barbarahui, line 53: Why not use `os.path.join` here? From my understanding after having read about `os.sep.join` vs `os.path.join`, the latter is more intelligent. You could end up with 2 separators in a row when using `os.sep.join`.

See https://docs.python.org/3/library/os.html?highlight=os%20sep#os.sep

content_harvester/by_page.py
```diff
@@ -0,0 +1,397 @@
...
+
+        if os.path.exists(children_path):
+            child_pages = [file for file in os.listdir(children_path)
+                        if file.startswith(parent_id)]
+            for child_page in child_pages:
+                child_page_path = os.sep.join([children_path, child_page])
```
barbarahui, line 59: Same question here re `os.sep.join`

content_harvester/by_page.py
```diff
@@ -0,0 +1,397 @@
...
+            f"[{collection_id}, {page_filename}, {calisphere_id}] "
+            f"Media Path: {media_dest}"
+        )
+
+        # Harvest Thumbnail File for Record
+        thumbnail_src = record.get('thumbnail_source')
+
+        # this makes it so we don't have to re-write isShownBy mappings for
+        # non-nuxeo sources
+        if isinstance(thumbnail_src) == str:
```
barbarahui, line 228: Shouldn't this be `if isinstance(thumbnail_src, str)` ?

content_harvester/by_page.py
```diff
@@ -0,0 +1,397 @@
...
+            'filename',
+            list(filter(lambda x: bool(x), src_url.split('/')))[-1]
+        )
+
+        tmp_file_path = os.path.join('/tmp', filename)
+        if os.path.exists(tmp_file_path):
+            return tmp_file_path
+
+        # Weird how we have to use username/pass to hit this endpoint
+        # but we have to use auth token to hit API endpoint
```
barbarahui, line 290: Yeah, hrm, I don't know if this is the case anymore since we upgraded Nuxeo and got rid of Shibboleth.

content_harvester/derivatives.py
```diff
@@ -0,0 +1,95 @@
...
+    magick_location = CONTENT_PROCESSES['magick']
+    process = [
+        magick_location, "-quiet", "-format", "-jp2", "-define",
+        "jp2:rate=10", f"{source_file_path}[0]", jp2_path
+    ]
+    msg = f"{source_file_path} converted to {jp2_path}"
+
+    subprocess.check_output(process, stderr=subprocess.STDOUT)
+    print(msg)
+    return jp2_path
```
barbarahui, line 95: So much simpler! 👍

barbarahui (changes requested):
OK, I looked through it all on my own and it makes sense! Added a few inline comments/requests.

metadata_mapper/mappers/nuxeo/nuxeo_mapper.py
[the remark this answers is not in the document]
amywieliczka: replaced `'url': url.replace('/nuxeo/', '/Nuxeo/')` with `'url': file_content.get('data', '').strip()` and removed line 203.

barbarahui (approved):

PR #337, opened 2023-03-03
1. Implement Islandora child mappers

lthurston:
Things to discuss:

**caltech:** 

- How do we handle "restricted," which is basically a mechanism for skipping some records based on some condition? Might want to think about this holistically, solving the occasional need to skip some records based on conditions that are unique to the mapper.
- Not tested at all because Caltech endpoints are not working.

**sjsu:**

-  It does not modify its parent class. Does this seem right?


lthurston:
[] Remove SJSU
[] Look into using Vernacular subclasses for the skipping mechanism mentioned above

lthurston:
This is still a work in progress, but I've removed SJSU, and have implemented an extremely basic skipping mechanism.

lthurston:
This still needs to be tested, and compared with Solr once the caltech endpoints are working.

lthurston:
@amywieliczka @barbarahui There are only a couple mappers after the others got voted off the island: burbank and chs. They're working and this PR is ready for review.

amywieliczka (approved):

lthurston:
@amywieliczka Had a merge conflict that I rebased out, so I need another approval. Sorry!

amywieliczka (approved):

amywieliczka:
@lthurston should be all set.

PR #338, opened 2023-03-09
2. Simplify Arck mapper, allow mapping to Callables in the process

lthurston:
This is not ready for review. Further changes to the mapper approach are necessary to prevent parent mappers from being called during the build process.

### DIFF
--- a/metadata_mapper/__init__.py
+++ b/metadata_mapper/__init__.py
@@ -1,0 +1,1 @@
+import utilities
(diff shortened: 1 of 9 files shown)


lthurston:
@amywieliczka @barbarahui This is ready for review. It started as a simplification of Arck, but ended up with changes to when mapper helper methods and mapper methods are invoked.

lthurston:
There's a little issue here with imports that I need to sort. Setting this back to draft.

amywieliczka (approved):

amywieliczka (approved):

amywieliczka (approved):

PR #350, opened 2023-03-13
3. Implement set of mappers for March

lthurston:


lthurston:
I need to make a little change to PSPL based on this:

> Hi 
>[@lucas](https://castiron.slack.com/team/U024ZDD2Q)
> following on our meeting this morning, and the issue you uncovered with Palm Springs Public Library's `is_shown_by` values resulting in a 404 with the message: `Error: No ‘key’ parameter was provided.` We reached out to them, and a key contact at Veridian software responded: the URL syntax for the thumbnails changed, from this (where Chia1949.1.1 is variable -- comprising the unique identifier for the item, from within the metadata):
> http://collections.accessingthepast.org/cgi-bin/imageserver.pl?oid=Chia1949.1.1&width=200&ext=jpg
> to this:
> https://collections.accessingthepast.org/?a=is&oid=Chia1949.1.1&type=pagethumbnailimage&width=200

metadata_mapper/mappers/oai/omeka/nothumb_mapper.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class NothumbVernacular(OmekaVernacular):
+    record_cls = NothumbRecord
+
+    def skip(self, record):
+        """
+        This is a lightweight version of OmekaRecord.map_is_shown_by(). Any changes here may need to be
+        reflected there.
+        """
```
amywieliczka, line 15: Thanks for the code comment. It's not great, but I agree with your decision to leave it like this for now. If this becomes a regular issue, then maybe we should re-think how we handle skips.

amywieliczka (dismissed):

lthurston:
@amywieliczka Ready for final approval now!

amywieliczka (approved):

PR #351, opened 2023-03-15
Indexer

barbarahui:


barbarahui:
@amywieliczka I've opened this PR as a draft because there's a lot that still needs to be figured out and improved in terms of the architecture. 

The basic functionality of creating the OpenSearch cluster exists in `record_indexer/cloudformation/`

The basic functionality of creating a index template for rikolti indices exists in `index_collection.py`

The basic functionality for adding a collection to the index exists in `indexer.py`

Some things to figure out:

- Do we run this in Lambda and batch up collections into smaller chunks so we don't hit the 15 minute limit, or should we use Fargate? How much data do we want to hold in memory?
- Naming: I appreciate your point about the verb `to index` being overloaded and thus potentially confusing, so maybe I can find more specific verb(s) to replace it.
- I think this needs a similar architectural solution to the content fetcher re. running by collection vs by registry endpoint vs by single object. I didn't have time to think about elegant ways to do this yet.

I suppose this might qualify as "baseline" development and so could be merged in and close issue #292 ? On the one hand it doesn't feel great to have unpolished code merged into main, but on the other hand it can get messy to leave code hanging out for a long time. Curious what you think.

record_indexer/index_templates/rikolti_template.py
```diff
@@ -0,0 +1,240 @@
...
+        "Content-Type": "application/json"
+    }
+
+    # create index template
+    r = requests.put(url, headers=headers, data=json.dumps(payload), auth=AUTH)
+    r.raise_for_status()
+    print(r.text)
+
+
+def get_properties():
```
amywieliczka, line 56: given that this just returns a static JSON structure, does this really need to be a function? I feel like you've maybe overthought this a bit.

record_indexer/settings.py
```diff
@@ -0,0 +1,31 @@
...
+
+AWS_ACCESS_KEY_ID = os.environ.get('AWS_ACCESS_KEY_ID', False)
+AWS_SECRET_ACCESS_KEY = os.environ.get('AWS_SECRET_ACCESS_KEY', False)
+AWS_SESSION_TOKEN = os.environ.get('AWS_SESSION_TOKEN', False)
+AWS_REGION = os.environ.get('AWS_REGION', False)
+
+if not LOCAL_RUN and DATA_SRC == 'local':
+    print(
+        "A local data source or local data destination "
+        "is only valid when the application is run locally"
```
amywieliczka, line 19: what's the local data destination?

record_indexer/settings.py
```diff
@@ -0,0 +1,31 @@
+import os
+
+from dotenv import load_dotenv
+
+load_dotenv()
+
+LOCAL_RUN = os.environ.get('INDEXER_LOCAL_RUN', False)
```
amywieliczka, line 7: what's a local run? what's not a local run?

amywieliczka (changes requested):

record_indexer/delete_index.py
```diff
@@ -0,0 +1,26 @@
+import sys, os
+import settings
+import requests
+import argparse
+
+ENDPOINT = os.environ.get('RIKOLTI_ES_ENDPOINT')
+AUTH = ('rikolti', os.environ.get('RIKOLTI_ES_PASS'))
```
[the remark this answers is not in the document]
amywieliczka: I've moved these into settings. I've also updated all use of ENDPOINT to use fstrings, so `url=f"{settings.ENDPOINT}/{index}"` - they shouldn't use `os.path.join` because they're not operating system paths, they're urls, which universally use a `/` character and don't need the operating system interface provided by the builtin `os` module!

record_indexer/index_templates/rikolti_template.py
```diff
@@ -0,0 +1,240 @@
...
+import copy
+
+'''
+    Create OpenSearch index template for rikolti
+    https://www.elastic.co/guide/en/elasticsearch/reference/7.9/index-templates.html
+    https://www.elastic.co/guide/en/elasticsearch/reference/7.9/indices-component-template.html
+'''
+
+ENDPOINT = os.environ.get('RIKOLTI_ES_ENDPOINT')
+AUTH = ('rikolti', os.environ.get('RIKOLTI_ES_PASS'))
```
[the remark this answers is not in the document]
amywieliczka: See commit 911551ccdd57e87851037cddd4e7fce74ba0523c

record_indexer/index_templates/rikolti_template.py
amywieliczka: See commit ad082dd3fc87bd64ea08c959629a048b0822ea0e

record_indexer/settings.py
amywieliczka: Removed in commit 911551ccdd57e87851037cddd4e7fce74ba0523c - we can add it back if it becomes useful.

amywieliczka:
Hey @barbarahui - I've made a few updates:
- I moved the variables `ENDPOINT` and `AUTH` into settings.py - that way we only ever get `ENDPOINT` and `AUTH` from the environment once, which simplifies the interface between this python component and its environment. 
- I removed `settings.LOCAL_RUN` for the time being, since everything is run on local hardware. In the fetcher and mapper, this flag determines whether the fetcher and mapper call the lambda functions, or use boto to invoke a new serverless lambda function instance. 
- I flake8'ed the whole thing and fixed dead simple flake8 errors
- I created a json file for the index template, rather than using python to build the basically static data structure from other data structures. 
- I chose to put the entire payload into the `record_index_config.json` file, but you could instead have a `record_schema_config.json` file that contains `{"title": {"type: "text", "fields": {"raw": {"type": "keyword"}}}, "alternative_title": {"type": "text"...}...}` and you could keep the `settings` block of this data structure in `rikolti_template.py`. I don't really follow why you had it broken out the way you did. 
- I also chose to include the `"children": {"type": "nested"}` key/value pair in the json file, but I do dynamically modify it in `rikolti_template.py` to add a "properties" key rather than copying the entire schema in the json file. 

Please review these updates and let me know what you think. I think lambda-fying this should be relatively straightforward, though maybe seems a bit dumb since there is, at this point, exactly one record per file. But I do think you've got your `by_collection.py` and `lambda_function.py` mixed up in order of magnitude.

record_indexer/cloudformation/create_cf_template_opensearch.py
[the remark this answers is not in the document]
barbarahui: I removed `cf_opensearch.yaml` from the repo. It shouldn't have been checked in in the first place.

I also added instructions to the README on how to create the OpenSearch cluster, which should clarify the relationship between the scripts. See https://github.com/ucldc/rikolti/pull/351/commits/d65ddb776378c885bdcba237e8de6bc38687a02a

record_indexer/delete_index.py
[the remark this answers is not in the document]
barbarahui: Thanks for making these changes. Totally makes sense to import from settings. And also yeah, no reason at all to use `os.path.join` here!

record_indexer/index_templates/rikolti_template.py
barbarahui: OK yeah, I see your point. I was using python to generate the JSON for me, but there's no need to memorialize that in the repo. It's easier to see what's going on here.

record_indexer/settings.py
barbarahui: Right, of course, we don't need to set `LOCAL_RUN` here because the lambda doesn't call itself as is.

record_indexer/settings.py
barbarahui: There is no local destination, only source. When I was doing development, I didn't want to keep on hitting S3 to retrieve the json files to be indexed. The json files were on my local file system, and I was adding them to an index hosted on AWS.

barbarahui:
I reviewed your changes and responded to comments throughout.

As far as your last point about `by_collection.py` and `lambda_function.py` being mixed up in order of magnitude: well, I was trying to start thinking about batching up records rather than having them indexed one at a time. Like I said in my comment at the top of the page, I didn't have time to really think about this yet. What is/was in the repo right now is definitely a draft, which was why I originally submitted it as a draft PR and asked your opinion on whether or not it should be an official PR.

I agree with you that Lambda is maybe not a good solution here. I put some notes in the README [here](https://github.com/ucldc/rikolti/blob/indexer/record_indexer/README.md#index-records) and am planning to return to this question.

PR #352, opened 2023-03-22
Implement PRA fetcher

lthurston:
BLOCKED by:

- CDL: discovery regarding the varying response types
- CDL: direction regarding number of requests per invocation

This fetcher is very rough, requires more documentation, but works on the two collections I've been testing it against. It should be considered a proof of concept at this point.

I've added some comments to the code which you can see in the [Files changed tab](https://github.com/ucldc/rikolti/pull/352/files). There are two main things I wanted to call your attention to:

1. The first request, the `original_url`, is composed of the internal collection id and a hardcoded URL, and returns a `ChildrenResponse`, which can be composed of structural objects or information objects. The former requires additional requests to get to a list of information objects. I've provided specific examples in my code comments, alongside the code that deals with these two scenarios. The primary question I have about this is whether or not this is expected, or if there's an oddball collection in our midst.
2. This fetcher, unlike some of the others, requires additional requests to get the item metadata from the list of paginated results. Because of this, I needed to add an additional, optional method to the Fetcher class that can be overridden in the children when necessary. I called it `get_text_from_response`. In the Fetcher class, you'll see that this method is quite simple. In the PRA fetcher, it's a bit more involved.

I'll add that because so many requests are needed to get the metadata, this fetcher is much slower than the others I've used to date. Two requests take place for each item in the result set, so 100 items can take a a minute or two.

Please take a look when you get a chance and we can discuss on Monday.

### DIFF
--- a/metadata_fetcher/fetchers/preservica_api_fetcher.py
+++ b/metadata_fetcher/fetchers/preservica_api_fetcher.py
@@ -1,0 +1,239 @@
+import json
+from xml.etree import ElementTree
+from .Fetcher import Fetcher
+from requests.auth import HTTPBasicAuth
+from requests.adapters import Retry
+from requests.adapters import HTTPAdapter
+import requests
+import re
+
+
+class PreservicaApiFetcher(Fetcher):
+    BASE_URL: str = "https://us.preservica.com/api/entity/v6.0"
+
+    NAMESPACES: dict = {
+        "pra": "http://preservica.com/EntityAPI/v6.0",
+        "xip": "http://preservica.com/XIP/v6.0"
+    }
+
+    def __init__(self, params: dict[str]):
+        """
+        Parameters:
+            params: dict[str]
+        """
+        super(PreservicaApiFetcher, self).__init__(params)
+
+        # Tracks where we're at processing record sub-requests
+        self.record_index = 1
+        self.record_total = 0
+
+        # If `next_url` is a param, we know that this is not
+        # the fetch of the first page, so skip setting those
+        # attributes
+        if "next_url" in params:
+            for key in params:
+                setattr(self, key, params[key])
+            return
+
+        credentials = params.get("harvest_data").get("harvest_extra_data")
+        self.basic_auth_credentials = ([v.strip() for v in credentials.split(',')])
+        self.internal_collection_id = \
+            re.search(
+                r"(?<=SO_)[0-9a-f-]+",
+                params.get("harvest_data").get("url")
+            ).group(0)
+
+        self.next_url = (
+            f"{self.BASE_URL}/structural-objects/"
+            f"{self.internal_collection_id}/children?max=1"
+        )
+
+    def build_fetch_request(self) -> dict[str]:
+        """
+        Generates arguments for `requests.get()`.
+
+        Returns: dict[str]
+        """
+        return self.build_url_request(self.next_url)
+
+    def aggregate_vernacular_content(self, response: str) -> str:
+        """
+        TODO: at time of dev, response is a requests.Response, but will be str when
+        merged
+
+        Parameters:
+            response: str
+
+        Returns: str
+        """
+
+        response_body = response.text
+
+        # Starting with a list of `information-objects` URLs
+        object_url = ElementTree.fromstring(response_body).\
+            find("pra:Children/pra:Child", self.NAMESPACES).text
+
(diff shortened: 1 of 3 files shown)


metadata_fetcher/fetchers/Fetcher.py
```diff
@@ -78,29 +78,34 @@ def fetch_page(self):
...
 
+
         if self.check_page(response):
+            text = self.get_text_from_response(response)
             if settings.DATA_DEST == 'local':
-                self.fetchtolocal(response.text)
+                self.fetchtolocal(text)
             else:
-                self.fetchtos3(response.text)
+                self.fetchtos3(text)
```
lthurston, line 87: Because the metadata for each item is not fetched from the same URL that provides the paginated list of items (by reference) and two additional further HTTP requests are required to fetch the metadata for the page, an additional method was added to allow the generation of a pseudo-response that contains the paginated list with the item metadata. This method `get_text_from_response`, by default, will return `resolve.text`, see below.

metadata_fetcher/fetchers/pra_fetcher.py
```diff
@@ -0,0 +1,159 @@
...
+        if len(io_children) > 0:
+            return self.original_url
+
+        child_url = root.find(".//pra:Child[@type='SO']", self.NAMESPACES).text
+
+        request = self.build_url_request(child_url)
+        response = requests.get(**request)
+        root = ElementTree.fromstring(response.text)
+
+        return root.find(".//pra:AdditionalInformation/pra:Children", self.NAMESPACES).text
```
lthurston, line 61: Further discovery is required, but the `original_url`, the first URL generated from the internal collection id seems to come in at least two forms, which can be see in looking at this URL for collections:

26782: `https://us.preservica.com/api/entity/v6.0/structural-objects/079822c0-b1fc-45df-8166-19880b12edba/children`

26460: `https://us.preservica.com/api/entity/v6.0/structural-objects/fd56c609-a9aa-4e8e-8745-60d1128befe0/children`

The first one lists children with the type of IO, information objects. This is a page of results, with the <Child> elements referencing the objects and their metadata.

The second lists a single child of the type SO, a structural object, which requires an additional request to get to the paginated list of items.

The fetcher currently works for both these collections, but needs to be tested against others.

metadata_fetcher/fetchers/pra_fetcher.py
```diff
@@ -0,0 +1,159 @@
...
+
+        # Getting an individual `information-object`, extracting the URL
+        # for the oai_dc metadata fragment
+        metadata_urls = {object_url: self.get_metadata_url_from_object(object_url)
+                         for object_url in object_urls}
+
+        # Getting the metadata
+        items = {object_url: self.get_metadata_from_url(metadata_url)
+                 for (object_url, metadata_url) in metadata_urls.items()
+                 if metadata_url is not None}
```
lthurston, line 87: Here I'm gathering the metadata into a single result file, starting from the response from the request for paginated items.

metadata_fetcher/fetchers/pra_fetcher.py
```diff
@@ -0,0 +1,159 @@
...
+
+        # Replace XML text node with the response
+        output_document = response.text
+
+        for search, replace in items.items():
+            output_document = output_document.replace(search, replace)
+
+        return output_document
```
lthurston, line 95: This is not how I intend to construct to fetched page of data. Consider it more of a proof of concept.

aturner:
@lthurston we checked in with Preservica re: their caveat that Basic Authentication (using the API) may not be supported long-term. They have no immediate plans to remove it, but it's evidently been deprecated for multiple releases and they reserve the right to remove at any point. Their suggestion, below -- can we update the fetcher to authenticate with a token?

==

The recommended way to authenticate the APIs is via the auth token, there is a API call which will return a token which can be used to make API calls.

Authentication API Documentation (preservica.com)
https://developers.preservica.com/blog/getting-started-with-preservica-access-tokens

The python SDK uses these access tokens and manages the process for you.
https://pypreservica.readthedocs.io/en/latest/intro.html#authentication

lthurston:
@aturner I can take a look, sounds like a good plan. Are we still waiting on an answer to the fetcher / mapper question?

barbarahui:
Closing. Superseded by https://github.com/ucldc/rikolti/pull/1333

PR #379, opened 2023-03-29
1. Implement flickr fetcher

lthurston:
Requires the addition of `FLICKR_API_KEY` env var.

It would be easy to modify this to perform the minimum number of requests per invocation, meaning 2 (one for the list, one for the individual photo record).

### DIFF
--- a/metadata_fetcher/fetchers/flk_fetcher.py
+++ b/metadata_fetcher/fetchers/flk_fetcher.py
@@ -116,7 +116,9 @@ class FlkFetcher(Fetcher):
         """
         return {
             "api_key": settings.FLICKR_API_KEY,
-            "method": method
+            "method": method,
+            "format": "json",
+            "nojsoncallback": "1"
         }
 
     def build_request_url(self, params: dict[str]) -> str:
@@ -147,21 +149,20 @@ class FlkFetcher(Fetcher):
     def get_text_from_response(self, response: requests.Response) -> str:
         """
         Accepts a response from a request for page of photos, and transforms it
-        in a dictionary. This requiresa `flickr.photos.getInfo` request for each
-         photo.
+        in a dictionary. This requires a `flickr.photos.getInfo` request for
+        each photo.
 
         Parameters:
             response: requests.Response
 
         Returns: str
         """
-        set_element_tree = ElementTree.fromstring(response.content)
-        photos = set_element_tree.findall(".//photo")
+        photos = json.loads(response.content)
 
         photo_data = []
-        for photo in photos:
-            content = self.get_photo_metadata(photo.attrib["id"]).content
-            photo_data.append(self.flickr_photo_info_xml_to_dict(content))
+        for photo in photos.get("photos", {}).get("photo", []):
+            content = self.get_photo_metadata(photo.get("id")).content
+            photo_data.append(json.loads(content).get("photo"))
             self.photo_index += 1
 
         print(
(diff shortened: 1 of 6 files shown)


lthurston:
@timfrazee I'd love your feedback on this before I share with the CDL team.

metadata_fetcher/fetchers/flk_fetcher.py
```diff
@@ -0,0 +1,311 @@
...
+
+    @staticmethod
+    def get_api_key() -> str:
+        """
+        Fetches the Flicker API key. This approach avoids storing it in state
+        params.
+
+        Returns: str
+        """
+        return os.environ.get('FLICKR_API_KEY', 'boguskey')
```
timfrazee, line 45: For consistency, this should fetch the key from the settings module.

lthurston:
@amywieliczka @barbarahui This is ready for review. The one thing that I haven't thought too hard about is the change in `metadata_fetcher/lambda_function.py` to handle the change to the fetcher name. Let me know what you'd like me to do there.

metadata_fetcher/fetchers/flk_fetcher.py
```diff
@@ -0,0 +1,257 @@
...
+    def get_current_url(self) -> str:
+        """
+        Returns the request URL based on the type of collection, photoset or
+        user_photos. Note we increment the page number since Flickr starts
+        with page 1 and internal page iteration starts with page 0.
+
+        Returns: str
+        """
+        if self.is_photoset():
+            return self.get_photoset_request_url(self.harvest_extra_data,
```
barbarahui, line 44: Can you move the implementation of `self.get_photoset_request_url()` into this function instead of making it a separate function?

metadata_fetcher/fetchers/flk_fetcher.py
```diff
@@ -0,0 +1,257 @@
...
+                                                 self.write_page + 1)
+        return self.get_user_photos_request_url(self.harvest_extra_data,
```
barbarahui, line 46: Can you move the implementation of `self.get_user_photos_request_url()` into this function instead of making it a separate function?

metadata_fetcher/fetchers/flk_fetcher.py
```diff
@@ -0,0 +1,257 @@
...
+        Returns: bool
+        """
+        data = json.loads(http_resp.content)
+        self.photo_total = len(data.get(self.response_items_attribute, {}).
+                               get("photo", []))
+
+        print(
+            f"[{self.collection_id}]: Fetched page {self.write_page} "
+            f"at {http_resp.url} with {self.photo_total} hits"
+        )
```
barbarahui, line 209: Can you change this to:
```
        print(
            f"[{self.collection_id}]: Fetched ids for page {self.write_page} "
            f"at {http_resp.url} with {self.photo_total} hits"
        )
```

metadata_fetcher/fetchers/flk_fetcher.py
```diff
@@ -0,0 +1,257 @@
...
+        in a dictionary. This requires a `flickr.photos.getInfo` request for
+        each photo.
+
+        Parameters:
+            content: str
+
+        Returns: str
+        """
+        photos = json.loads(content)
+
```
barbarahui, line 162: Can you add:
```
        print(
            f"[{self.collection_id}]: Starting to fetch all photos for page"
            f" {self.write_page}"
        )
```

barbarahui (changes requested):
This method of making multiple requests for a page of fetched metadata looks good! We've made some comments inline. As a general principle, we're trying to avoid doing transformation of the vernacular metadata in the fetcher, so if you can rename `transform_vernacular_content()` to `aggregate_vernacular_content()` as per the comment, that would be great.

metadata_fetcher/lambda_function.py
```diff
@@ -8,6 +8,9 @@
 
 
 def import_fetcher(harvest_type):
+    if harvest_type == "flickr":
+        harvest_type = "flk"
```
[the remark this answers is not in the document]
lthurston: I figured as much. I'm going to leave it for now so I can continue fetching, but will remove when we're ready to merge.

lthurston:
@barbarahui All changes have been made except the flickr -> flk change. Added retries too, although not in a particularly sophisticated way. Am happy to adjust however you see fit.

amywieliczka (dismissed):
Looks good!

Not sure what the overhead is for creating a new `requests.Session()` for fetching each object vs. having a single self.session that's created in in the `__init__` function and persists alongside the lifespan of the `FlkFetcher`. Might be worth looking into as a next step for retries - adding it as a card to the project board.

barbarahui (approved):

metadata_fetcher/lambda_function.py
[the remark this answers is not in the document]
lthurston: @amywieliczka do you have a better solution than this workaround I have in place?

lthurston:
I pushed this branch again after rebasing on top of main. However, I think we're waiting until we have a solution for my little hack in the lambda_function, which is an unresolved conversation.

metadata_fetcher/lambda_function.py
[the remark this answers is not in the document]
lthurston: This is resolved.

lthurston:
This PR is ready to be merged.

barbarahui (approved):

PR #384, opened 2023-04-06
1. flickr fetcher / mappers

lthurston:
- The description values with the Rikolti mappers is a modest improvement over the legacy mapper results. The source data is quite bad, resulting from using the description field as a metadata store in inconsistent ways.
- There's an issue with validation of collection 27125 that we need to look into.


lthurston:
@barbarahui @amywieliczka Despite all my huffing and puffing about the `sdasm` mapper, I think I got it in a fairly good place. I have yet to validate against solr, and want to do that before I say this is officially ready for review. If you have time to take a look at these mappers to see if you have any questions, comments, objections, that'd be great. If not that's cool, too.

The multiple executions of the `split_description()` method on `sdasm` and `sppl` are not the most egregious waste of electricity / CPU cycles ever (that dubious award goes to cryptocurrency mining of course), but it does give me a minor twinge in the part of my brain that seeks efficiency. Like Elsa, I can let it go.

metadata_mapper/mappers/mapper.py
```diff
@@ -639,13 +639,14 @@ def enrich_subject(self):
...
+
         if isinstance(subjects, str):
             subjects = [subjects]
         if not subjects:
             subjects = []
         subjects = [
-            subject.get('name') for subject in subjects
-            if isinstance(subject, dict)
+            subject.get('name') if isinstance(subject, dict) else subject
+            for subject in subjects
```
amywieliczka, line 649: Rather than changing this in the enrichment, could you update the mapper to make subject conform to this `[{'name': <subject>}]` data structure? 

I know it seems strange to build this data structure only to have an enrichment remove the data structure, but this algorithm change here in the enrichments would require some data analysis across every collection running the `enrich_subject` enrichment.

metadata_mapper/mappers/mapper.py
```diff
@@ -1110,11 +1111,16 @@ def capitalize_value(self, exclude):
             "sourceResource/contributor",
             "sourceResource/extent",
             "sourceResource/format",
+            "sourceResource/type",
```
amywieliczka, line 1114: why was this added here?

amywieliczka (changes requested):

metadata_mapper/mappers/flickr/flickr_mapper.py
```diff
@@ -0,0 +1,85 @@
...
+class FlickrRecord(Record):
+    def UCLDC_map(self):
+        return {
+            "calisphere-id": self.legacy_couch_db_id.split('--')[1],
+            "isShownAt": self.map_is_shown_at,
+            "isShownBy": self.map_is_shown_by,
+            "date": self.map_date,
+            "description": self.map_description,
+            "format": ["Photo"],
+            'identifier': [self.source_metadata.get("id"), self.url_image],
```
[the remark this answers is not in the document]
lthurston: I'm not sure either. Some of the flickr-fetched mappers do have values like this in the Solr data:

```                        'identifier': ['40982751',
                                       '14_001411.tif',
                                       '14_001411'],
```

But that's just a filename, not a URL, so I'm having trouble explaining this one. Additionally, that first value also doesn't match the `id` value, which, in this case, is `c9a8add342ed62e287e53f18c5b633f9`. The oddball single quotes are giving debugging / troubleshooting vibes too, but I don't know how I got there. Apologies!

I will take a deeper look the identifier values on these flickr mappers and try to figure this out.

metadata_mapper/mappers/flickr/sdasm_mapper.py
```diff
@@ -0,0 +1,142 @@
...
+
+        # Set the description if it wasn't provided as metadata in the
+        # description field
+        if "description" not in description_parts:
+            description_parts.update({"description": description})
+
+        return description_parts
+
+    def map_date(self):
+        return list(filter(None, [
```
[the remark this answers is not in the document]
lthurston: I think that would work fine. I will make that change.

metadata_mapper/mappers/mapper.py
lthurston: Sure thing.

metadata_mapper/mappers/mapper.py
lthurston: I'm sorry, I need to take better notes on changes like this next time. The explanation for this completely escapes me. I'll remove it and run this through the validator.

lthurston: It appears that the `type` field is lowercased, seemingly by another enrichment because the mapper returns the value "Image" with the first letter capitalized. My change was a way to make it match the value in Solr.

lthurston: It's the `enrich_type` enrichment that's doing it. Perhaps this is intentional, solr is wrong and the type should be lowercase.

metadata_mapper/mappers/flickr/flickr_mapper.py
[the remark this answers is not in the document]
lthurston: I removed the `self.url_image` bit. This configuration only impacts `flickr.flickr`, which is a little unreliable to validate at the moment. I'm going to close this because this specific issue is resolved, but have switch this back to draft until we can successfully validate each of the mappers.

amywieliczka (approved):

PR #388, opened 2023-04-14
Set up docker dev env

lthurston:
@amywieliczka and @barbarahui, I took a little of my own time and Dockerized this project, basically making it so developer environments can be run in a container. I updated the README with instructions, it's quite simple. It primarily facilitates what I have to do a lot when developing locally.

Here are the top line takeaways:

- Using ipython
- Can start console in either the fetcher or mapper module
- Modules inside the cwd are autoloaded when the console starts
- Use `autoreload` to reload all loaded modules

I've tested this, but haven't done any development with it, so I've yet to see how much of an improvement it is, but not having to deal with pyenv, the easy reloading, plus everything else that comes with ipython should make things a bit easier.

Downsides:

- There's a hardcoded background color in ipython that can't be changed, I had to configure my terminal to remap it to something sane (after pulling my hair out for a while), see: https://github.com/ipython/ipython/issues/13446
- There's an intermittent issue where, for example, after reloading while working on a mapper I'll get an error that the vernacular class doesn't extend Vernacular, requiring a restart of the console.

We haven't discussed this, but I thought you might be interested in trying it out to see how you like it.

Also: I haven't dug into it yet, but starting `bin/map` does result in an error right now, which might indicate something that needs to be fixed in the main branch.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,19 @@
+name: Continuous Integration
+on:
+  pull_request:
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.9"
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install ruff
+      - name: Run Ruff
+        run: ruff check --format=github .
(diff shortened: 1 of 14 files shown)


lthurston:
I added a Github action to link with Ruff on PRs.

amywieliczka (approved):

barbarahui (approved):
Looks good!

amywieliczka:
Okay to merge even though linter checks have failed.

PR #403, opened 2023-04-21
Validator improvements

lthurston:
@timfrazee I'm putting validator tweaks, improvements here.


lthurston:
@amywieliczka @barbarahui I'd like to get these little fixes approved soon if possible. I think we'll open another PR for further validation work.

PR #407, opened 2023-04-27
Use rikolti_mapper_type instead of mapper_type

amywieliczka:
@lthurston, if you've made any changes to this mapper_type lookup map, please let me know. I haven't had much time to make updates to the collection registry, so it's unlikely those changes have been propagated through, and this work here will break things.

### DIFF
--- a/metadata_mapper/map_registry_collections.py
+++ b/metadata_mapper/map_registry_collections.py
@@ -5,41 +5,6 @@ import sys
 import lambda_shepherd
 import logging
 
-# TODO: this sort of translation from registry's mapper_type to
-# rikolti's mapper_type should really be done in the registry.
-# once we have a firmer rikolti vocabulary of mappers, we should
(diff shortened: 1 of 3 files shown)


lthurston (approved):

metadata_mapper/map_registry_collections.py
```diff
@@ -76,11 +41,11 @@ def map_endpoint(url):
             ))
             logging.debug(log_msg.format(f"lambda payload: {collection}"))
             try:
-                collection['mapper_type'] = lookup[collection['mapper_type']]
+                # collection['mapper_type'] = lookup[collection['mapper_type']]
```
barbarahui, line 44: We can remove this line, right?

barbarahui (approved):
Looks good! Just one minor inline comment re removing a line of old code that was commented out.

PR #416, opened 2023-05-09
Fetcher local recursion

amywieliczka:
Review bug-fixes PR first

This branch implements some better reporting on fetching tasks:
- `check_page` returns the number of records found on the fetched page, rather than a boolean - updated for the 4 existing fetchers; @lthurston please indicate if this is a problem moving forwards
- `fetch_page` method returns the number of records found on the fetched page, as reported by the `check_page` method
- `lambda_function.fetch_collection` has a more thorough and detailed return value
- `lambda_function.fetch_collection` extends the return value for all pages when run recursively in the local environment
- `fetch_registry_collections.fetch_endpoint` makes use of this return value to provide details about how many records were fetched over how many pages

### DIFF
--- a/metadata_fetcher/fetch_registry_collections.py
+++ b/metadata_fetcher/fetch_registry_collections.py
@@ -9,7 +9,7 @@ import logging
 def fetch_endpoint(url, limit=None):
 
     collection_page = url
-    results = []
+    results = {}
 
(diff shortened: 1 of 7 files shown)


barbarahui (approved):
Looks good to me!!

PR #421, opened 2023-05-17
Validator Improvements Mk. II

### LINKED ISSUE #383: Validation testing for isShownAt + isShownBy from CouchDB

### LINKED ISSUE #386: Create a hook to create data validation report after mapping is complete
Could this be a metadata_mapper lambda_shepherd and lambda_fetcher payload flag? That would give workflow orchestration the opportunity to determine if it wanted to get a validation report. 

For now, by default, the metadata_mapper does not generate a validation report.

### LINKED ISSUE #411: Validator reports: the commas in the 'description' fields tab text to the next cell
Attached is an example from the validation report for collection #26981.

Image

### LINKED ISSUE #412: Validator reports: additional options to surface a) "content match" summary row; and b) "content match" field details
Here's a mock-up of what we discussed:

Image

- Default base report: Reports errors/warns (this is done)
- Toggle option A: INFO summary line, listing out fields with "content match"
- Toggle option B: INFO details, listing out all "content match" fields and values

timfrazee:
Enhances the mapping validator with additional functionality:

### Validator

* Reorganizes validator classes into their own modules within the metadata_mapper package
  * `ValidationErrors` is now `ValidationLog`
* Adds `ValidationLogLevel`
  * This enum is used to set the logging level
  * Options are `ERROR`, `WARNING`, `INFO`, `DEBUG`
* Adds `ValidationMode`
  * An enum `STRICT` or `LAX`
    * `STRICT` is the current default behavior
    * `LAX` performs case-insensitive string comparisons and order-insensitive list comparisons
  * Individual validations can define their own `validation_mode` as either `ValidationMode.STRICT` or `ValidationMode.LAX` to override the global default.
* Adds `verbose` option

* Adds successful validation logging at the `INFO` level
  * By default, outputs a single line per record, listing the successfully-validated fields
  * If `verbose` is True, outputs a line per successfully-validated field. WARNING: This can make for very long validation files.

### validate_mapping.py

* Fetches `isShownAt` and `isShownBy` from CouchDB and merges it into comparison metadata
* Added CLI args:
  * `--log-level`, which can be any one of `ERROR`, `WARN`, `INFO`, or `DEBUG` (case insensitive) (default: WARN)
  * `-v` `--verbose` to set verbose mode (default: false)
  * `--filename` to set the output file name (default: timestamp-based name)

For example:

```bash
python validate_mapping.py 12345 --log-level=INFO --verbose --filename=example.csv
```
will generate some console output telling you what's happening, and create a file in the env-appropriate bucket under "validation/{collection_id}/{filename}"

### Lambda Shepherd
* Added ability to trigger post-mapping validation by passing `validate` in the Lambda payload
  * `validate` can be a truthy boolean to run with default options, `False` to explicitly not validate, or a `dict` of options
  * The options dict can contain the same arguments as the CLI flags: `log_level`, `verbose`, and/or `filename`

Resolves #383
Resolves #386
Resolves #411 
Resolves #412

### DIFF
--- a/metadata_mapper/mappers/mapper.py
+++ b/metadata_mapper/mappers/mapper.py
@@ -12,7 +12,8 @@ from typing import Any, Callable
 from datetime import date
 
 import settings
-from .validator import Validator, ValidationErrors
+from validator.validator import Validator
+from validator.validation_log import ValidationLog
 
 from . import constants
 from .iso639_1 import iso_639_1
@@ -139,7 +140,7 @@ class Record(ABC, object):
         return {}
 
     # Validation
-    def validate(self, comparison_data: dict) -> ValidationErrors:
+    def validate(self, comparison_data: dict) -> ValidationLog:
         return self.validator.validate(self.mapped_metadata, comparison_data)
 
     # Mapper Helpers
--- a/metadata_mapper/validate_mapping.py
+++ b/metadata_mapper/validate_mapping.py
@@ -8,14 +8,19 @@ from typing import Type
 import settings
 import utilities
 
-from mappers.validator import Validator
+from validator.validator import Validator
+from validator.validation_mode import ValidationMode
+from validator.validation_log import ValidationLogLevel
 
 urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
 
 
 def validate_collection(collection_id: int,
                         validator_class: Type[Validator] = None,
-                        validator: Validator = None
+                        validator: Validator = None,
+                        default_validation_mode = ValidationMode.STRICT,
+                        log_level = ValidationLogLevel.WARNING,
+                        verbose = False
                         ) -> Validator:
     """
     Validates all pages of a collection of mapped data.
@@ -35,7 +40,9 @@ def validate_collection(collection_id: int,
         validator_class = get_validator_class(collection_id)
 
     if not validator:
-        validator = validator_class()
+        validator = validator_class(default_validation_mode=default_validation_mode,
+                                    log_level = log_level,
+                                    verbose = verbose)
 
     for page_id in utilities.get_files("mapped_metadata", collection_id):
         validate_page(collection_id, page_id, validator)
@@ -99,10 +106,10 @@ def validate_page(collection_id: int, page_id: int,
 
 def create_collection_validation_csv(collection_id: int) -> None:
     result = validate_collection(collection_id)
-    result.errors.output_csv_to_bucket(collection_id)
+    result.log.output_csv_to_bucket(collection_id)
 
 
-## Private
+## Private-ish
 
 
 def get_mapped_data(collection_id: int, page_id: int) -> dict:
--- a/metadata_mapper/validate_mapping.py
+++ b/metadata_mapper/validate_mapping.py
@@ -18,7 +18,7 @@ urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
 def validate_collection(collection_id: int,
                         validator_class: Type[Validator] = None,
                         validator: Validator = None,
-                        default_validation_mode = ValidationMode.STRICT,
+                        validation_mode = ValidationMode.STRICT,
                         log_level = ValidationLogLevel.WARNING,
                         verbose = False
                         ) -> Validator:
@@ -32,6 +32,12 @@ def validate_collection(collection_id: int,
             The validator class to use. Can be derived if not provided.
         validator: Validator (default: None)
             A validator instance to use. Can be derived if not provided.
+        validation_mode: ValidationMode (default: ValidationMode.STRICT)
+            The validation mode to use (unless overridden in individual validations)
+        log_level: ValidationLogLevel (default: ValidationLogLevel.WARNING)
+            The lowest log level that should be included in the output
+        verbose: bool (default: False)
+            Include verbose output in file?
 
     Returns: Validator
         The validator containing errors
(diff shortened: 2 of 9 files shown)


metadata_mapper/validator/validator.py
```diff
@@ -239,17 +82,46 @@ def set_validatable_fields(self, fields: list[dict[str, Any]] = [],
 
         Returns list[dict]
         """
```
amywieliczka, line 84: Looks like `merge` is no longer an option here - can't actually add a review comment on line 71 because there's been no change on line 71.

amywieliczka (changes requested):

amywieliczka:
Hey Tim! This generally looks great. 

I left a few inline comments - mostly housekeeping. The biggest comment is about the usage of the word "default" - I think maybe we could find some better language for the module overall regarding the various different kinds of "defaults" there are in the module, but also some documentation write-up on module usage and module structure would likely go a long ways to address some of the confusion created by this overloading of the word "default". You've drafted some really nice PR messages for the validator work and I wonder if you could draft a Rikolti wiki update (github.com/ucldc/rikolti/wiki) with some Validator usage documentation.

metadata_mapper/mappers/mapper.py
```diff
@@ -11,7 +11,8 @@
 from datetime import date
 
 import settings
-from .validator import Validator, ValidationErrors
+from validator.validator import Validator
```
barbarahui, line 14: I think it makes sense to assume that we are using the `mapper.py` module as an import instead of running it directly as a script, which case we can change this to:

`from validator import validator`

metadata_mapper/mappers/mapper.py
```diff
@@ -11,7 +11,8 @@
...
+from validator.validation_log import ValidationLog
```
barbarahui, line 15: Similarly, change this to:

`from validator import validation_log` 

or even collapse the 2 lines into:

` from validator import validator, validation_log`

barbarahui: Then when you instantiate the classes from these modules, like on line 89, you can do:

`validator = validator.Validator`

metadata_mapper/validator/__init__.py
barbarahui, line 1: I don't think this file is necessary.

metadata_mapper/validate_mapping.py
```diff
@@ -1,78 +1,26 @@
...
-from typing import Type, Union
+from typing import Type
 
 import settings
 import utilities
 
-from mappers.validator import Validator, ValidationErrors
+from validator.validator import Validator
+from validator.validation_mode import ValidationMode
+from validator.validation_log import ValidationLogLevel
```
barbarahui, line 13: Again, I think this can just be:

`from validator import validator, validation_mode, validation_log`

barbarahui (changes requested):
Looks good, thanks Tim! Mostly comments about `__init__.py` and import statements.

metadata_mapper/mappers/mapper.py
amywieliczka: Hmm, I do like using classes without having to specify a module prefix, especially when it's not at all confusing where the class came from, as in this case.

barbarahui: Hmm yeah I see what you're saying. Well, maybe we can just leave this as is for now and circle back when we choose a workflow management tool and need to figure out the best approach to packaging up rikolti.

metadata_mapper/validate_mapping.py
```diff
@@ -125,44 +75,165 @@ def validate_page(collection_id: int, page_id: int,
...
+    ret = {}
+
+    for field_name in ["isShownAt", "isShownBy"]:  
+      url = "https://harvest-prd.cdlib.org/" \
+            "couchdb/ucldc/_design/all_provider_docs/" \
+            "_list/has_field_value/by_provider_name_wdoc" \
+            f"?key=\"{collection_id}\"&field={field_name}&limit=1000"
+      
+      #### TODO: REMOVE verify=False once cert issues are resolved!
+      response = requests.get(url, verify=False)
```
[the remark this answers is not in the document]
timfrazee: Thanks for the reminder. Done!

metadata_mapper/validator/validator.py
timfrazee: Cleaned up the docstring to better reflect the new simplified functionality of this method.

metadata_mapper/validate_mapping.py
```diff
@@ -1,78 +1,26 @@
...
-    vernacular = utilities.import_vernacular_reader(mapper)
-
-    return vernacular.record_cls.validator
-
-
 def validate_collection(collection_id: int,
                         validator_class: Type[Validator] = None,
-                        validator: Validator = None
+                        validator: Validator = None,
+                        default_validation_mode = ValidationMode.STRICT,
```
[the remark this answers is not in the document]
timfrazee: Agreed. I cleaned up the argument names and updated the method docstring.

amywieliczka (dismissed):

barbarahui (approved):

timfrazee:
Hi @amywieliczka and @barbarahui, I rebased on `main` and force pushed - no other changes, but it should now merge nicely.

When you're able, please re-approve and/or go ahead and merge it.

Thanks!

amywieliczka (approved):

PR #436, opened 2023-06-07
Validator improvements

lthurston:
This PR includes:

- A quick and dirty way to convert couchdb keys into snake case for easy comparison. Amy is going to revisit the names of these fields so, this may only be necessarily in the short term.
- Tweak to validator to use the specified validator class on the vernacular when present, or fall back to Validator.

### DIFF
--- a/metadata_mapper/validate_mapping.py
+++ b/metadata_mapper/validate_mapping.py
@@ -4,6 +4,7 @@ import sys
 import urllib3
 
 from typing import Type
+from re import sub
 
 import settings
 import utilities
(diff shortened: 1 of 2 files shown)


metadata_mapper/validate_mapping.py
```diff
@@ -180,11 +191,11 @@ def get_couch_db_data(collection_id: int,
             "_list/has_field_value/by_provider_name_wdoc" \
             f"?key=\"{collection_id}\"&field={field_name}&limit=1000"
       
-      response = requests.get(url)
+      response = requests.get(url, verify=False)
```
amywieliczka, line 194: afaik this should no longer be necessary? I thought @barbarahui fixed this.

amywieliczka (approved):

metadata_mapper/validate_mapping.py
barbarahui: Yeah, the ssl cert is renewed and valid. We should remove `verify=False` from this line. 

I think we also need to add `certifi` to `requirements.txt` in order to avoid a separate error being thrown by `requests` though. See: https://stackoverflow.com/questions/22027418/openssl-python-requests-error-certificate-verify-failed

PR #437, opened 2023-06-07
Enrichment and tooling adjustments

lthurston:
There are a few commits here that address enrichment and tooling issues that came up during mapping development. The extended commit comments offer context.

### DIFF
--- a/metadata_mapper/lambda_shepherd.py
+++ b/metadata_mapper/lambda_shepherd.py
@@ -26,7 +26,7 @@ def check_for_missing_enrichments(collection):
 
     not_yet_implemented = []
     collection_enrichments = (
-        collection.get('rikolti__pre_enrichments', []) +
+        (collection.get('rikolti__pre_enrichments') or []) +
         collection.get('rikolti__enrichments')
     )
     for e_url in collection_enrichments:
--- a/metadata_mapper/mappers/mapper.py
+++ b/metadata_mapper/mappers/mapper.py
@@ -1121,6 +1121,10 @@ class Record(ABC, object):
             "dataProvider",
             "provider/name"
         ]
+
+        def capitalize_str(value):
+            return f"{value[0].upper()}{value[1:]}" if len(value) > 0 else ""
+
         for field in props:
             if field in exclude:
                 continue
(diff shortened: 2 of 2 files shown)


metadata_mapper/lambda_shepherd.py
```diff
@@ -26,8 +26,8 @@ def check_for_missing_enrichments(collection):
 
     not_yet_implemented = []
     collection_enrichments = (
-        collection.get('rikolti__pre_enrichments', []) +
-        collection.get('rikolti__enrichments')
+        (collection.get('rikolti__pre_enrichments') or []) +
+        (collection.get('rikolti__enrichments') or [])
```
amywieliczka, line 30: Pre-enrichments are optional, but all collections should have enrichments. What collection(s) didn't have enrichments?

metadata_fetcher/fetch_registry_collections.py
```diff
@@ -27,7 +27,16 @@ def fetch_endpoint(url, limit=None):
...
                 break
+
             log_msg = f"[{collection['collection_id']}]: " + "{}"
+
+            if not collection["ready_for_publication"]:
+                print(log_msg.format(
+                    f"Skipping collection {collection['collection_id']} "
+                    "(not ready for publication)"
+                ))
+                continue
```
amywieliczka, line 38: I really think this should actually just get added to the registry url that you provide as input to this script. For example: https://registry.cdlib.org/api/v1/rikoltifetcher/?format=json&mapper_type=up_oai_dc&ready_for_publication=true

The whole point of the harvester is to fetch and map collections not ready for publication. Then, when they have been QAed and we are certain the mappings are correct and the contributing institution is happy with how their records look, we mark them as ready for publication and push the records into our published index. 

In your case 'ready for publication = False' is doubling as a 'not yet ready for harvesting' flag. Collections can wind up in this state for all kinds of reasons, even while the Collection record still persists in the Collection Registry. 

I think it makes more sense to add `ready_for_publication=true` to the API URL passed to this script in those cases where you want to skip testing certain collections.

amywieliczka (changes requested):

metadata_mapper/lambda_shepherd.py
lthurston: Good question. IIRC it was a collection with a non-iterable value for `rikolti__pre_enrichments`. I made it consistent here, but an error if enrichments are missing altogether makes sense.

I'll change it to:

```
(collection.get('rikolti__pre_enrichments') or []) +
(collection.get('rikolti__enrichments'))
```
if that makes sense to you.


If you want me to, I will seek out the collection with the unexpected pre-enrichment value. It must have been Flickr, Solr or OAC.

amywieliczka (approved):

metadata_mapper/lambda_shepherd.py
amywieliczka: @lthurston I just went through Flickr and it's not Flickr (https://registry.cdlib.org/api/v1/rikolticollection/?format=json&harvest_type=flk), my guess is that it's OAC? 

Enrichments were previously unstructured text blocks. I spent awhile trying to parse them correctly into a structured and standardized array. I'm sure my regex didn't catch every case, but at the time, we had over 20,000 collections in the collection registry with many different flavors of malformed enrichment text blocks. I'd appreciate it if you do manage to track down the collection with the unexpected pre-enrichment value. If not, I'm working my way through running my fetcher-mapper-validator over everything and will probably encounter it then. Though maybe with this code in place, it won't actually error out.

lthurston: I will track it down!

lthurston: Did this attribute name get changed from to `rikolti__pre_enrichments` to `rikolti__pre_mapping`?

PR #453, opened 2023-06-28
Validation skips reporting out full pages of "new" or "missing" records

amywieliczka:
Lines 92-97 return early if the entire page is not found in either Rikolti or Solr, but this means that full pages of new Rikolti records (not found in Solr) and full pages of missing Solr records (not found in Rikolti) are never reported, and that new Rikolti records or missing Solr records are only reported if they sit next to at least one other matched record on the page. 

We definitely want to know if Rikolti harvests 210 records for a collection that has only 110 records in Solr, though! This may be valid, but should be reported and escalated to a person to evaluate what is happening.

### DIFF
--- a/metadata_mapper/validate_mapping.py
+++ b/metadata_mapper/validate_mapping.py
@@ -89,12 +89,12 @@ def validate_page(collection_id: int, page_id: int,
                         context=context
                       )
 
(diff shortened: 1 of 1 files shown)


barbarahui (approved):
Makes sense and looks good!

PR #481, opened 2023-08-07
UCD fetcher / mapper

lthurston:
Included:

- [x] Resolve issue with calisphere-id in UCLDC_map()
- [x] Resolve issue with legacy couch db id
- [x] Decide how to render calisphere id when missing the ark
- [x] Properly write to local or s3 after recent refactor
- [x] Add type annotations
- [x] Create a general function for fetching URLs w/ retries (`make_http_request()`) in the Fetcher base class

### DIFF
--- a/metadata_fetcher/fetchers/ucd_json_fetcher.py
+++ b/metadata_fetcher/fetchers/ucd_json_fetcher.py
@@ -47,11 +47,11 @@ class UcdJsonFetcher(Fetcher):
         ns = {"ns": "http://www.sitemaps.org/schemas/sitemap/0.9"}
         xml = ElementTree.fromstring(response.text)
         loc_nodes = xml.findall(".//ns:loc", ns)
-        pages = math.ceil(len(loc_nodes) / self.per_page) - 1
+        pages = math.ceil(len(loc_nodes) / self.per_page)
 
         for page in range(pages):
-            print(f"[{self.collection_id}]: Fetching URLs for page {page} "
-                  f"({page + 1}/{pages + 1})")
+            print(f"[{self.collection_id}]: Fetching URLs for page {page + 1} "
+                  f"({page + 1}/{pages})")
             skip = self.write_page * self.per_page
             urls = loc_nodes[skip:(skip + self.per_page)]
             records = list(filter(None, [self.fetch_json_ld(url.text) for url in urls]))
--- a/metadata_fetcher/fetchers/ucd_json_fetcher.py
+++ b/metadata_fetcher/fetchers/ucd_json_fetcher.py
@@ -22,6 +22,7 @@ class UcdJsonFetcher(Fetcher):
 
     def fetch_page(self) -> int:
         """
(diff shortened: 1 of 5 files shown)


lthurston:
Docker changes have been removed from this branch.

christinklez:
Hi @lthurston! We've reviewed the validation reports for the UCD fetcher/mapper set.
Only linking to one particular validation report, for [Collection #8](https://docs.google.com/spreadsheets/d/1WtWtx1FYlCuosQei-eyLsCckMdrL2AiGcU8-VYq7hQI/edit#gid=1478019524), but the issues listed below showed up in many other collections. (Let me know if you'd like to see the other reports, but note that the linked #8 report has the most bit of annotated notes for these issues. Thank you!

- **title & description:** We expect the data to have brackets wrapping around the text; Rikolti is not wrapping brackets. (Is this a mapping issue?) See examples in rows 2 & 7. Please wrap the data in brackets
- **rights / rights_uri:** We expect the rights info to map to rights_uri; it is currently mapping to rights. Please map to rights_uri. See examples in rows 4 & 6. 
- **title & rights -- type mismatch:** Type mismatch errors--expect "list of str" vs actual "str". Please investigate type mismatch errors.
- **subject:** Subject terms are not mapping. Subject terms seem to appear in the "about" fields in the vernacular data. See vernacular example in row 8. Please map subject terms.
- **publisher:** 20 records in this collection are incorrectly mapping the publisher data. mapped metadata is showing: "publisher": ["name"]. We expect to see what's currently in the expected value ("UCD Library..."). See example (as well as vernacular data variations) in row 15653. Please investigate publisher mapping anomaly (in 20 records only; other records are okay.)

lthurston:
@christinklez I've addressed these issues in this PR.

metadata_fetcher/fetchers/ucd_json_fetcher.py
```diff
@@ -0,0 +1,92 @@
...
+            params: dict[str]
+        """
+        super(UcdJsonFetcher, self).__init__(params)
+
+        self.collection_id = params.get("collection_id")
+        self.url = params.get("harvest_data").get("url")
+        self.per_page = 10
+
+    def fetch_page(self) -> int:
+        """
```
amywieliczka, line 24:
```suggestion
        """
        UCD's harvesting endpoint gets us an XML document listing a URL for every record in a collection, but not the actual metadata records themselves. fetch_page fetches this initial XML document, while fetch_all_pages below batches the records in sets of 10 (self.per_page) and fetch_json_ld fetches individual records.
```

metadata_mapper/lambda_function.py
```diff
@@ -102,13 +102,13 @@ def map_page(collection_id: int, page_filename: str, collection: Union[dict, str
     mapped_records = [record.solr_updater() for record in mapped_records]
     mapped_records = [record.remove_none_values() for record in mapped_records]
 
+    group_page_exceptions = {}
```
amywieliczka, line 105: Thank you

amywieliczka (approved):
Looks good to me. Just added a code comment suggestion

PR #493, opened 2023-08-17
chore: Restructures imports based on isort

bibliotechy:
Updates the the structure of import statements for all python files in metadata_fetcher, metadata_mapper, and content_harvester.

It follows the basic structure used by [isort](https://pycqa.github.io/isort/index.html) which breaks imports into sections and alphabetizes them within that section. Sections are
*  FUTURE
* STDLIB
* THIRDPARTY
* FIRSTPARTY
* LOCALFOLDER

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,13 @@ We use PR reviews to approve or reject, comment on, and request further iteratio
 - DRY
 - Readability & Transparency: Code as language
 - Favor explicitness over defensiveness
+- Import statements grouped according to [isort](https://pycqa.github.io/isort/index.html) defaults:
+  - FUTURE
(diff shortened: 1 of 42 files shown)


amywieliczka (approved):

barbarahui (approved):
As I said in my inline comment, I'm honestly unclear as to whether or not the relative import statements will work. I tried running one or 2 pieces of code in a local env and got import errors, but maybe my setup isn't correct and/or I'm running the code from the wrong directory. According to the documentation I've read online, it *should* work, so I'm going to approve this PR since I know it's holding up the subsequent PR, and figure it out later.

PR #502, opened 2023-08-25
Barbara/airflow mapper

barbarahui:
@amywieliczka @bibliotechy Can you take a look at this draft of this mapper dag and give feedback on the approach? Any and all feedback very welcome!

Some specific questions/comments: 

1. It seems like parameters can only be accessed inside a task run, meaning that params needs to be unpacked both in `fetch_pages()` and then again in `map_page()` (or passed from one task to another) which seems clunky. Chad, are you aware of a different approach to this?
2. Amy, can you verify that this is the format of the payload we're expecting to get from Registry? (lines 57-61)
3. Do we have any collections that have > 1024 pages of vernacular metadata? I'm guessing that some of the Nuxeo complex object collections exceed that. Assuming so, I'll need to figure out how to handle this case. (See my comments on lines 69-72). We can of course increase the value to > 1024, but that seems like a recipe for overloading our MWAA instance.

### DIFF
--- a/dags/mapper_dag.py
+++ b/dags/mapper_dag.py
@@ -49,12 +49,8 @@ def map_page_task(page: str, collection: dict, params=None):
     if not collection_id:
         return {}
 
-    payload = {'collection_id': collection_id}
-    payload.update({'page_filename': page})
-    payload.update({'collection': collection})
-
     try:
-        mapped_page = map_page(json.dumps(payload))
+        mapped_page = map_page(collection_id, page, collection)
     except KeyError:
         print(
             f"[{collection_id}]: {collection['rikolti_mapper_type']} "
@@ -67,7 +63,7 @@ def map_page_task(page: str, collection: dict, params=None):
     schedule=None,
     start_date=datetime(2023, 1, 1),
     catchup=False,
-    params={'collection_id': None},
+    params={'collection_id': Param(None, description="Collection ID to map")},
     tags=["rikolti"],
 )
 def mapper_dag():
--- a/dags/mapper_dag.py
+++ b/dags/mapper_dag.py
@@ -1,11 +1,9 @@
 from datetime import datetime
 import json
-from time import sleep
-
-import requests
 
 from airflow.decorators import dag, task
 from airflow.models.param import Param
+
 from rikolti.metadata_mapper.lambda_shepherd import get_vernacular_pages, get_collection
 from rikolti.metadata_mapper.lambda_function import map_page as lambda_map_page
 
@@ -17,7 +15,6 @@ def fetch_pages(params=None):
 
     payload = params.get('payload')
     collection_id = payload.get('collection_id', 0)
-
     # raise an error?
     if not collection_id:
         return []
@@ -36,7 +33,7 @@ def map_page(page: str, params=None):
     collection_id = payload.get('collection_id', 0)
     # raise an error?
     if not collection_id:
-        return []
+        return {}
 
     collection = get_collection(collection_id)
     payload.update({'collection': collection})
--- a/dags/mapper_dag.py
+++ b/dags/mapper_dag.py
@@ -1,5 +1,6 @@
 from datetime import datetime
 import json
+import sys
 
 from airflow.decorators import dag, task
 from airflow.models.param import Param
--- a/dags/mapper_dag.py
+++ b/dags/mapper_dag.py
@@ -14,8 +14,7 @@ def get_vernacular_pages_for_collection_task(params=None):
     if not params:
         return False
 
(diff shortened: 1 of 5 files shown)


dags/mapper_dag.py
```diff
@@ -0,0 +1,77 @@
...
+
+    payload = params.get('payload')
+    collection_id = payload.get('collection_id', 0)
+    # raise an error?
+    if not collection_id:
+        return {}
+
+    collection = get_collection(collection_id)
+    payload.update({'collection': collection})
+    payload.update({'page_filename': page})
```
amywieliczka, line 41: yikes, yeah I think we want this to be different.

amywieliczka (changes requested):

dags/mapper_dag.py
```diff
@@ -0,0 +1,77 @@
+from datetime import datetime
+import json
+import sys
+
+from airflow.decorators import dag, task
+from airflow.models.param import Param
+
+from rikolti.metadata_mapper.lambda_shepherd import get_vernacular_pages, get_collection
+from rikolti.metadata_mapper.lambda_function import map_page as lambda_map_page
```
amywieliczka, line 9: I think we should try to move away from the use of the word "lambda" in our codebase. Changing the filenames `lambda_shepherd` and `lambda_function` feels like a big shift, but let's not add `lambda_map_page`? The fetcher uses `fetch_page_task` to differentiate between the airflow task and the function `fetch_page` which is doing the work of fetching the page.

bibliotechy:
@barbarahui @amywieliczka My biggest outstanding question is around how we structure the full workflow, from fetching the vernacular to publishing the updated collection to stage calisphere.

If each "part" of the workflow (fetch, map, content fetch, etc) is its own dag, how will we build a single workflow where an operator can trigger a dag with a single collection ID and have the entire workflow execute?

amywieliczka:
@bibliotechy agreed - I think a `harvest_collection` dag makes sense, which is comprised of the full pipeline. Though also, I don't think what's done here poses a major problem, and for development iteration purposes, it might be nice to have an isolated component up until the point that it is considered done? 

My understanding with the taskflow API is that you can treat tasks like functions for use in multiple dags, and I think even import them between different dag definition python modules. I don't think that's particularly ideal, though.

barbarahui:
@bibliotechy If we put all of the tasks in a single DAG so that it's possible to run a harvest from start to finish, how do we allow users to run just a single task, e.g. if somebody wants to just rerun the mapper? I guess we add a parameter that allows the user to specify which task(s) to run, and then we have an initial task that parses this and somehow passes that back to the DAG, which runs only the tasks specified? Is that the right model? Like Amy, I'm still a bit puzzled about airflow parameters.

dags/mapper_dag.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+@task()
+def fetch_pages(params=None):
+    if not params:
+        return False
+
+    payload = params.get('payload')
+    collection_id = payload.get('collection_id', 0)
```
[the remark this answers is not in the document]
barbarahui: Yeah, let's just standardize on one way of doing it?

dags/mapper_dag.py
```diff
@@ -0,0 +1,77 @@
...
+    # raise an error?
+    if not collection_id:
+        return []
```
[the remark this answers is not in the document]
barbarahui: I set the default value to "None" for the mapper dag. That seems to make sense to me. Let's discuss and standardize on a way to handle this too.

dags/mapper_dag.py
```diff
@@ -0,0 +1,77 @@
...
+
+    pages = get_vernacular_pages(
+                collection_id)
```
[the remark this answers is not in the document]
barbarahui: I went ahead and factored out the rollup reporting from `lambda_shepherd.map_collection` into `lambda_shepherd.get_mapping_summary` so that it can be called from the mapper dag. Lmk what you think.

dags/mapper_dag.py
barbarahui: I refactored the code so that only `collection_id` is required as input for running the mapper. Eager to hear your feedback.

barbarahui: After all of the refactoring, `lambda_shepherd` and `lambda_function` are definitely misnamed! I made the assumption that we no longer need lambda entry points, so payload input isn't supported by the code anymore. (Let me know if there's some reason we should maintain this after all).

How about this renaming:
lambda_shepherd.py --> map_collection.py
lambda_function.py --> map_page.py

amywieliczka (approved):

PR #514, opened 2023-09-18
Address oai.samvera validation issues

lthurston:


amywieliczka:
I'm confused why there's so much happening in this mapper but there was so little happening in the legacy mapper: https://github.com/calisphere-legacy-harvester/dpla-ingestion/blob/cfe3dcb06008c0c6cb9d8207fe28bfaa1a855e4f/lib/mappers/ucsc_oai_mapper.py#L6 Are there things that should be happening in the base OAI mapper class that aren't???

lthurston:
@christinklez @aturner @amywieliczka @barbarahui 

I'm attaching a screenshot and a full CSV of the validation report from collection 154 after fixing a few things. There are some questions:

- Content mismatch on identifier: the `uclamss_1387_*` values don't show up in the vernacular, so this feels like data drift. I do see similar strings in filenames, but it would require transformation to generate these.
- Content mismatch on is_shown_at: URL domain was ursus.library.ucla.edu, now is digital.library.ucla.edu. I can write a validation rule to ignore this if that's helpful.
- Content mismatch on date: there are two date formats provided, mapping both seems correct to me, but I can modify the logic for this mapper if desired
- Content mismatch on subject: different order of list items

Screenshot 2023-11-13 at 2 17 38 PM
[11-13-2023T21-54-23.csv](https://github.com/ucldc/rikolti/files/13342629/11-13-2023T21-54-23.csv)

aturner:
@lthurston In looking at what we've previously harvested from the old fetcher/mapper vs. Rikolti results, it looks like the first 3 items can be attributed to data drift.

UCLA has been migrating content across new systems, so the change to the item landing page/edm:isShownAt URL makes sense. 

We are AOK with the multiple dates, and also the reordered subjects.

metadata_mapper/mappers/oai/samvera_mapper.py
```diff
@@ -3,12 +3,70 @@
...
+
     def UCLDC_map(self):
         return {
-            'is_shown_at': self.source_metadata.get('isShownAt'),
-            'is_shown_by': self.source_metadata.get('isShownBy')
+            "alternativeTitle": self.map_alternative_title,
+            "type": self.map_type,
         }
 
+    def map_alternative_title(self) -> list:
```
barbarahui, line 19: I think this function should be moved to the OAI mapper.

metadata_mapper/mappers/oai/samvera_mapper.py
```diff
@@ -3,12 +3,70 @@
...
+    def map_is_shown_by(self) -> str:
+        value = self.source_metadata.get("object")
+
+        if isinstance(value, list):
+            return value[0]
+        elif isinstance(value, str):
+            return value
+        return None
+
+    def map_type(self) -> list:
```
barbarahui, line 45: I think this function should also be moved to the OAI mapper. Or was there a reason to put it here? The logic is hairy in the legacy harvester, but it looks to me like it could be in the base class.

barbarahui, line 45: I think this function should also be moved to the OAI mapper. Or was there a reason to put it here? The logic is hairy in the legacy harvester, but it looks to me like it could be in the base class.

barbarahui, line 45: I think this function should also be moved to the OAI mapper. Or was there a reason to put it here? The logic is hairy in the legacy harvester, but it looks to me like it could be in the base class.

lthurston: Current OAI mapper is simpler:

`            "type": self.source_metadata.get("type")`

That said, I think you're right that the hairier logic is compatible with even simple values. I'll give this a go.

metadata_mapper/mappers/oai/samvera_mapper.py
lthurston: Will do.

amywieliczka (approved):

amywieliczka:
Marking this as ready for review and merging as these updates make sense for a re-inventory of OAI mappers.

PR #518, opened 2023-09-21
Inputs outputs logs

amywieliczka:
Most notably, this PR creates a `harvest_dag.py` and moves all dag tasks definitions there. Keeping code local to the general case - running the full harvesting pipeline - was helpful in thinking about structuring the inputs and outputs as ideal to this general case. @bibliotechy I can move these back either temporarily (to make reviewing the changes made by this PR within each task a bit easier), or permanently if you think it makes more sense to functionally encapsulate these task definitions. 

This PR also
- standardizes on using `params` method to get dag run parameters (as opposed to `dag_run.conf` or `get_current_context()`)
- standardizes dags as taking a collection id while dag tasks should take a collection metadata resource from the registry API. 
- updates the fetcher task to print the fetcher task status, and return a list of successfully fetched page filenames, eliminating the need to then go get the vernacular pages as the first step of the mapping task in the full harvest pipeline (though this is still needed when running just the mapper_dag, and this may require some finessing when using an s3 data source)
- standardizing on usage of the words "status" to refer to conceptual task status (number of pages successfully fetched, mapping enrichment errors encountered, etc), rather than "summary" and "report"

### DIFF
--- a/dags/fetcher_dag.py
+++ b/dags/fetcher_dag.py
@@ -7,6 +7,7 @@ from rikolti.dags.harvest_dag import get_collection_fetchdata_task
 from rikolti.dags.harvest_dag import fetch_collection_task
 
 @dag(
+    dag_id="fetch_collection",
     schedule=None,
     start_date=datetime(2023, 1, 1),
     catchup=False,
--- a/dags/fetcher_dag.py
+++ b/dags/fetcher_dag.py
@@ -3,7 +3,7 @@ from datetime import datetime
 from airflow.decorators import dag
 from airflow.models.param import Param
 
-
+from rikolti.dags.harvest_dag import get_collection_fetchdata_task
 from rikolti.dags.harvest_dag import fetch_collection_task
 
 @dag(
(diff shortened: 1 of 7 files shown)


amywieliczka:
@bibliotechy I'm confused about the Dags Check failure. I'm able to run this locally without any dag import errors in the airflow UI or the airflow local-runner container. 

This section on the Taskflow API suggests that you can import and reuse tasks definitions: https://airflow.apache.org/docs/apache-airflow/stable/tutorial/taskflow.html#reusing-a-decorated-task, but maybe I need to specify a specific-to-the-dag task id? I would have assumed tasks get stored as a dag-task, such that I could reuse them without having to differentiate them in a global namespace.

bibliotechy:
@amywieliczka I'm also confused by that. It looks like it is at the dag ID level, and not the task ID level that it is complaining. Maybe you can set a specific `dag_id` for each dag we define? But not sure why you would need to, as it sounds like it is supposed to pick up the id from the name of the function it is decorating, and they are different.

bibliotechy (approved):

PR #521, opened 2023-09-26
DAG loader mistakenly loads lambda_shepherd due to 'dag' in comments

amywieliczka:
Successfully resolves import error locally. 

Interestingly, even including "daguerreotype" in the comments triggers the DAG loader. This is definitely an Airflow bug.


bibliotechy (approved):
d🤯g

PR #523, opened 2023-09-27
Make some tentative changes to cleanup_value enrichment

lthurston:
This is not meant as a serious candidate for merging. Although it does make this enrichment modify the data, it likely will have side effects in other collections than the single one that I tested (3433, cca_vault). It seems unlikely that there's a one size fits all way for this enrichment to handle trailing periods in all contexts. More likely the mapper will need to handle some of this, and / or the cleanup_value() enrichment will need to relinquish its monopoly on record transformations.

Here are the changes:

1. Modify the field names to remove nested fields that don't exist in Rikolti
2. Rename variables so it's clearer what they do
3. Small changes to regexes to avoid IDE warnings
4. Changes to the fields that are operated on, both those that have a trailing do stripped and those that don't

Next steps might be to test this against more collections and have some discussions about how this enrichment could work better generally.

### DIFF
--- a/metadata_mapper/mappers/mapper.py
+++ b/metadata_mapper/mappers/mapper.py
@@ -1145,29 +1145,21 @@ class Record(ABC, object):
         """
         2076 times: no parameters
         """
-        default_fields = [
-            "sourceResource/language", "sourceResource/title",
-            "sourceResource/creator", "sourceResource/relation",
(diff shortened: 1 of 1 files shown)


lthurston:
This was a little exploratory PR, but not practical at the moment. Like Gordon Ramsay, I'm shutting it down.

PR #533, opened 2023-10-05
Content DAG

amywieliczka:
This is probably at a good point for review and evaluation. It pipes together a lot of the Dockerfile/docker-compose/mounting/DockerOperator/Airflow Variable work. 

### Docker/Airflow Integration Work:

The Content Harvester can be run as a python module (within a system with system level dependencies installed), as a docker container (recommend to use docker-compose), or as a docker operator within the `aws_mwaa_local_runner` ecosystem. This PR:

- Updates the Content Harvester's Dockerfile entrypoint to be congruent with the new import world order (`python -m content_harvester.by_page`)
- Sets up docker-compose for local development by overwriting the entrypoint, mounting `rikolti_data`, `rikolti_content`, and the source code into the container, and sets several environment variables for writing to these mounted volumes. 
- Creates a `ContentHarvesterDockerOperator` custom operator which subclasses the DockerOperator in order to package together standard DockerOperator parameters, make the `ContentHarvesterDockerOperator` importable in multiple DAGs, and make the `ContentHarvesterDockerOperator` available for dynamic fan-out with use of airflow's operator `.expand()` method. 
- `ContentHarvesterDockerOperator` uses airflow variables to specify content_harvester image name and content_harvester version (defaults to `content_harvester` and `latest`, respectively)
- `ContentHarvesterDockerOperator` manages figuring out mounts from the host machine (not the airflow container) to the content_harvester container, when run through the airflow container, via the `CONTENT_DATA_MOUNT` and `CONTENT_MOUNT` environment variables specified in `startup.sh`. This has been reflected in the README. 
- Creates a `dags/content_dag.py` containing two `ContentHarvesterDockerOperator`s, one operating to harvest content for a single page, one operating to harvest content for an entire collection (via entrypoint overwrite at the Operator level). 
- Alongside all this filesystem mounting work, there was work within the application code to update `settings.py` (and it's usage) to use the URL structure (ex: file:///path or s3://bucket/path) for `CONTENT_DATA_SRC`, `CONTENT_DATA_DEST`, and `CONTENT_DEST` environment variables. This has been reflected in the README. 

Features still requiring development work:
- TODO: had to add `chmod +x` permissions to `by_collection.py` to get that entrypoint working, but when I changed entrypoint from by_collection to by_page, I didn't update chmod - how is it working?
- TODO: there's the addition of a `get_mapped_page_filenames_task`, but this could possibly get folded into the `get_mapping_status_task`, which doesn't currently have a return value I believe. 
- deployment of content harvester container to ECR. 
- verification that the DockerOperator works as we expect in deployed MWAA context (we maybe probably need an airflow connection to the ECR repository? some roles to run the container? roles to write to s3 from the container?) with s3 `CONTENT_DATA_SRC`, `CONTENT_DATA_DEST` and `CONTENT_DEST` environment variables. 

### Content Harvester Application Work:
In addition to the changes to settings.py to get mounts all working correctly during local development, I also made the following updates to the content harvester application code: 

- https://github.com/ucldc/rikolti/pull/533/commits/f694506937090b5625dc7142bfb7b301a424ee5b - an abstraction refactor
- https://github.com/ucldc/rikolti/pull/533/commits/a370126e3b409c810488fa5344b811cb3e95fcd2 - move record-level reporting of missing thumbnails
- https://github.com/ucldc/rikolti/pull/533/commits/e9222087438d1b202b41f3bd8301de2798dc0449 - write per-page output instead of per record

I'm tempted to cherry-pick these out into a new branch dedicated to the internals of the content_harvester container. Along the way I ran into some application-level errors that were tricky to trace and wound up refactoring this codebase instead (first commit listed here). 

The second commit listed here is relatively straightforward: "Move record level reporting on missing thumbnails" allows us to report out missing thumbnails in the records that have been processed before an exception occurs and exits the process entirely. 

The third commit within the application code addresses the fact that the content_harvester was originally written to take a page of mapped metadata and spin off it's own lambda processes for each record. Each lambda process would read the record, find any described media files and/or thumbnail files, download them to the lambda's runtime environment, process them, and then upload the processed media and/or thumbnail files to their respective s3 locations before writing to s3 a revised mapped metadata record with details regarding where these files were written to. This 1:1 mapping of lambda process to record was determined due to lambda's 15 minute timeout, the indeterminate number of records per page, and the indeterminate amount of time all the network traffic of downloading from source institution and uploading to s3 might take. Since we no longer have those limitations, the third commit adjusts the code base to wait until a full page of mapped metadata records have been revised with media/thumbnail locations before writing a full page out, instead of writing out one file per record. The 1:1 file to page relationship still seems somewhat arbitrary, but this is consistent with the rest of our codebase. 

The contents of the content_harvester container still aren't quite fully featured. I started working on the internals of the content harvester some in the above commits, but then realized this PR is already looking pretty sizable. 

Features still requiring development work:
- md5 hashing of images
- s3 output
- cache optimizations

### Samples Files Isolating Airflow Concepts:
- Split up `taskflow_sample_dag` into a series of thematically related sample dags in `dags/samples/`. 
- `dags/samples/docker_debug.py` was helpful in debugging mounting the local docker socket into the airflow docker container. 
- `dags/samples/docker_operator.py` demos a simple airflow DockerOperator, with `dags/samples/simple_python_image` as the docker image. `dags/samples/docker_operator.py` also demos using jinja templating in the DockerOperator to access DAG parameters and mounting a folder from the host file system to the docker container that's run using the DockerOperator. Also demos creating a subclass of the DockerOperator for use in dynamic fan-out. 
- `dags/samples/dynamic_tasks.py` demos subclassing a BasicOperator for use in dynamic fan-out (extrapolated to a DockerOperator subclass in `dags/samples/docker_operator.py`)
- `dags/samples/airflowisms.py` demos getting an environment variable in the DAG definition and passing it into a task to validate that we can do this with the `CONTENT_DATA_MOUNT` and `CONTENT_MOUNT` environment variables. 
- `dags/samples/airflowisms.py` demos getting an airflow json variable in addition to a straightforward key:value variable. 
- `dags/samples/airflowisms.py` demos using an airflow variable in jinja templates with a BashOperator

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,7 +184,9 @@ export CONTENT_DEST=file:///rikolti_content
 
 The folder located at `CONTENT_DATA_MOUNT` is mounted to `/rikolti_data` and the folder located at `CONTENT_MOUNT` is mounted to `/rikolti_content` on the content_harvester docker container.
 
-Run `docker build -t content_harvester content_harvester` to build the `content_harvester` container locally.
+Run `docker build -t content_harvester content_harvester` to build the `content_harvester` container locally. 
+
+You can specify a `content_harvester_image` and `content_harvester_version` through the Airflow UI > Admin > Variables. The default value for `content_harvester_image` is `content_harvester` and the default value for `content_harvester_version` is `latest`.
 
 Finally, run `./mwaa-local-env build-image` to build the docker image, and `./mwaa-local-env start` to start the mwaa local environment.
 
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ vi env.local
 
 Currently, I only use one virtual environment, even though each folder located at the root of this repository represents an isolated component. If dependency conflicts are encountered, I'll wind up creating separate environments.
 
-Similarly, I also only use one env.local as well. Rikolti fetches data to your local system and then maps that data. Set `FETCHER_DATA_DEST` to the URI where you would like Rikolti to store fetched data - Rikolti will create a folder (or s3 prefix)`vernacular_metadata` at this location. Set `MAPPER_DATA_SRC` to the URI where Rikolti can find a `vernacular_metadata` folder that contains the fetched data you're attempting to map. Set `MAPPER_DATA_DEST` to the URI where you would like Rikolti to store mapped data - Rikolti will create a folder (or s3 prefix) `mapped_metadata` at this location.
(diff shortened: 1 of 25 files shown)


bibliotechy (approved):
OK, I have this running locally and I think the approach is fine. The setup still feels a bit complex, as we are defining the same value / env vars in multiple plces, but I don't have a solution for how to make it better, so I think we should move forward with this approach for now but keep an eye open for ways to improve it in the future.

PR #545, opened 2023-10-09
bugfix: no "records_mapped" key, use "count" instead

amywieliczka:


amywieliczka:
This really should have been a part of the inputs-outputs-logs work that was merged as part of #518

PR #549, opened 2023-10-10
fix: properly read and write to s3 from the metadata mapper

amywieliczka:


amywieliczka:
Merging this in without review in order to test functionality in mwaa.

PR #559, opened 2023-10-18
Don't overwrite type if registry's dcmi_type is None/X

amywieliczka:
will need to re-generate validation reports for flickr, cca_vault, and diff them against existing validation reports to ensure this doesn't create new errors.


amywieliczka:
@lthurston I think this is a duplicate of what you did in blacklight? can you confirm?

lthurston:
I based the Blacklight branch on this commit, so nothing was duplicated.

PR #560, opened 2023-10-19
[feat]: Validate by mapper type dag

amywieliczka:
This PR re-orders the `rikolti_data` folder, adds a `harvest.sh` and `harvest_collection.sh` scripts at the root of the directory (we can remove that `rikolti_bucket_template` branch now - a bucket template is no longer necessary), and adds a DAG to validate by mapper type which performs fetching, mapping, and validating tasks. 

The validate by mapper type dag can take a limit parameter (optional). If none is provided, it validates all collections of that mapper type. If a limit is provided, it validates up to the limit in the order the collections are provided by the registry api. 

This is fully functional but missing some obvious UI pieces, described below, as well as some scalability pieces - notes for myself as much as anyone else: 

Still required to get up and running:
- set up Adrian, Christine, and Gabriela with access to the rikolti_data s3 bucket and set up all tasks to output the s3 url of the rikolti_data OR
- make s3 bucket publicly accessible and set up all tasks to output the s3 url of the rikolti_data OR
- set up a copy s3 to google drive task (research the airflow operator, figure out how this should function)

I think the 3rd option is best, though I'm always skeptical about dealing with Google Drive API and authentication issues and I've not yet researched this. 

Next steps: 
- add an offset parameter to the validate by mapper type dag. 
- add a copy s3 to google drive task to move folders to google drive
- fix `rikolti.validator.validate` so it correctly adds missing records to the validation file instead of the job's log output, set up the validate_endpoint_task output the s3 url of the validation reports it's creating. 
- move solr and couch environment variables into airflow variables so Adrian, Christine, and Gabriela can easily and self-service switch between running validation jobs against solr/couch-stg and solr/couch-prd (how does this impact local development?)
- add dates to folders within the `rikolti_data` folder, ex: `rikolti_data/3433-2023_10_18/vernacular_metadata-3433-2023_10_18`, hook this in to <component>_data_src, <component>_data_dest options - maybe create a `RikoltiStorage` class that handles s3 vs. file storage generically that can be used throughout Rikolti - effectively wrapping boto s3 client. once a RikoltiStorage class is in place, we can define in one place how to assign dates (date + time for everything? just date? date + count - ie: `2023_10_18_run_1`, `2023_10_18_run_2`) and how to reference dates (always use most recent unless otherwise specified - what kind of hook to build in order to specify?)

### DIFF
--- a/dags/validate_by_mapper_type.py
+++ b/dags/validate_by_mapper_type.py
@@ -1,0 +1,67 @@
+import requests
+from datetime import datetime
+
+from airflow.decorators import dag, task
+from airflow.models.param import Param
+
+from rikolti.metadata_fetcher.fetch_registry_collections import fetch_endpoint
+from rikolti.metadata_mapper.map_registry_collections import map_endpoint
+from rikolti.metadata_mapper.map_registry_collections import registry_endpoint
+from rikolti.metadata_mapper.validate_mapping import create_collection_validation_csv
+
+@task()
+def make_mapper_type_endpoint(params=None):
+    if not params or not params.get('mapper_type'):
+        raise ValueError("Mapper type not found in params")
+    mapper_type = params.get('mapper_type')
+    return (
+        "https://registry.cdlib.org/api/v1/rikoltifetcher/?format=json"
+        f"&mapper_type={mapper_type}&ready_for_publication=true"
+    )
+
+@task()
+def fetch_endpoint_task(endpoint, params=None):
+    limit = params.get('limit', None) if params else None
+    return fetch_endpoint(endpoint, limit)
+
+@task()
+def map_endpoint_task(endpoint, params=None):
+    limit = params.get('limit', None) if params else None
+    return map_endpoint(endpoint, limit)
+
+@task()
+def validate_endpoint_task(url, params=None):
+    limit = params.get('limit', None) if params else None
+
+    response = requests.get(url=url)
+    response.raise_for_status()
+    total = response.json().get('meta', {}).get('total_count', 1)
+    if not limit:
+        limit = total
+
+    print(f">>> Validating {limit}/{total} collections described at {url}")
+
+    for collection in registry_endpoint(url):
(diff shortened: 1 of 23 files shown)


harvest.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+export RIKOLTI_HOME="/Users/awieliczka/Projects/rikolti"
```
barbarahui, line 3: Can we add `RIKOLTI_HOME` as an argument to the script? Or just instruct the user to set it in their env before running the script?

harvest_collection.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+export RIKOLTI_HOME="/Users/awieliczka/Projects/rikolti"
```
barbarahui, line 3: Same comment as for `harvest.sh`.

harvest.sh
barbarahui, line 1: It might be clearer to rename this harvest_by_mapper_type.sh or similar.

dags/validate_by_mapper_type.py
```diff
@@ -0,0 +1,76 @@
+import requests
+import logging
```
barbarahui, line 2: Lines 1 and 2 should be reversed if we're being picky :-)

barbarahui (changes requested):

harvest.sh
amywieliczka: I'm not trying to spend a whole lot of time on these `harvest.sh` and `harvest_collection.sh` scripts unless folks will use them. It seems like no one used the `rikolti_bucket_temp` branch that I threw up either. The branch has just been lingering, though, gradually getting more out of date, and my local repository has been all kinds of a mess with generating validation reports. This at least folds these two scripts in with the main branch and keeps things moving together, even though the contents of these scripts are hardcoded to be relevant to my machine. Do you think it's worth flushing this out into something more broadly functional here though?

barbarahui (approved):

PR #566, opened 2023-10-25
Validator type improvements

timfrazee:
* Replaces `type_match` validator class with `validate_type`, which takes a single arg of the expected type
* Removes `type` from validatable field definition
* Reimplements `type_match` as comparing the types of the rikolti and comparison/solr value
* General clean up and improvement of type comparison behavior


barbarahui:
@amywieliczka @timfrazee I don’t know exactly what the ask was here, but here are my observations after reading through the changes in this PR:

- It looks like the new `verify_type()` function could replace `type_match()`? If so, we'd just need to update the `sdasm_mapper` to use `verify_type()`. We could then also remove both the `type_match()` and `derive_type()` function, or keep both with a note that they are stubs.
- The signature for `add_validatable_field()` has changed. There are some subclassed Validators that need to be updated accordingly: sppl_mapper, sdasm_mapper, flickr_mapper.
- `add_validatable_fields()` is a new function as of this PR, but is already used in the `cca_vault_mapper`. I guess this means that Lucas's PR must have merged in this PR?

timfrazee:
@barbarahui 

- Yes, `verify_type` replaces `type_match`. I refactored `type_match` to perform a different job - it now compares the type of the Rikolti data against the type of the Solr data. This is not currently used anywhere, and could be removed if it's not useful. But it was easy to implement and seemed like it might be desirable if we were doing a lax content comparison (which casts values to strings) content comparison but still wanted to make sure the types matched.
- Yes, but only to remove `type`, which is no longer part of the validatable field definition, but instead part of the validation callable itself. I've updated the Validator subclasses that were present on the branch as of my last rebase. Before merging, we'd definitely want to rebase on main and update any others that have been introduced since then.
- In this branch, I updated `cca_vault_mapper`. If it's present on another branch, then I'd assume it was based on this one.

barbarahui:
@timfrazee Cool, then it looks good to me, pending making sure that subclasses are updated. Also pending @amywieliczka taking a look I think!

amywieliczka:
I count 3 direct subclasses of Validator, and 2 subclasses of FlickrValidator.

amywieliczka:
Completed this work & merging it in!

PR #567, opened 2023-10-25
create an order insensitive validation mode

amywieliczka:


amywieliczka:
This has already been merged as part of a different set of PRs

PR #623, opened 2023-11-07
Address validation issues in ucd_json

lthurston:
@timfrazee Check out this validator change when you get a chance.


amywieliczka:
Re-done here: https://github.com/ucldc/rikolti/pull/677

PR #628, opened 2023-11-15
Rikolti versioning

amywieliczka:
This PR introduces the `rikolti.utils.storage` and `rikolti.utils.versions` modules. 

`rikolti.utils.storage` is a module implementing list, get, and put operations for `s3://` or `file://` URIs. Currently, it implements the following functions, where a `data_uri` is always a URI formatted absolute path to a storage location and kwargs are always passed along to the underlying `boto3` call and can be used to add AWS credentials:
- `list_dirs(data_uri: str, recursive=False, **kwargs)`: returns a list of all directories that are the immediate child of `data_uri` - at this time, there is no recursive implementation. Returns relative paths - it's the only function in this module that does not deal in absolute paths, and it might be worth changing it for consistency, even though it's usage does not require absolute paths. 
- `list_pages(data_uri: str, recursive: bool=True, **kwargs)`: lists all files that are the child of `data_uri`; can recursively traverse a directory tree or return only direct children. Returns a list of URI-formatted absolute paths. 
- `get_page_content(data_uri: str, **kwargs)`: returns the contents of the file stored at `data_uri`. 
- `put_page_content(content:str, data_uri: str, **kwargs)`: writes `content` to a file located at `data_uri`, creates any subdirectories required to successfully write to the location described by `data_uri`, returns `data_uri`. 
- `upload_file(filepath:str, data_uri: str, **kwargs)`: moves the file located at `filepath` to `data_uri`, creates any subdirectories required to successfully write to the location described by `data_uri`, returns `data_uri`. 

The storage module deals exclusively in absolute path URI values, while `rikolti.utils.versions` deals primarily in "version paths" and "version pages", If `s3://bucket-name/3433/vernacular_data_v1/data/page1.xml` is the URI to the first page of vernacular data, then `3433/vernacular_data_v1/` is the "version" or "version path" and `3433/vernacular_data_v1/data/page1.xml` is the "version page". A version path always starts with the collection id. `rikolti.utils.versions` implements the creation of new version paths given an existing version path, a method to find a version path in an arbitrary string - usually an absolute path URI:
- `get_version(collection_id: Union[int, str], uri: str)`: takes an arbitrary string (usually a URI) and tries to find a version path by splitting on the collection id and discarding everything prior to the collection ID and also discarding everything after the special `data` keyword. returns a version path. 
- `create_version(base_version: str, pipeline_step: str, suffix: Optional[str] = None)`: takes a version path such as "3433/" or "3433/vernacular_metadata_v1", a pipeline step such as "vernacular_metadata" or "mapped_metadata", and a suffix, such as "v2", and produces a new version: "3433/vernacular_metadata_v2/" or "3433/vernacular_metadata_v1/mapped_metadata_v2". If no suffix is provided, uses the current datetime. 
- `create_vernacular_version`, `create_mapped_version`, `create_validation_version`, and `create_content_data_version` all call `create_version` while enforcing a particular name for the pipeline step. `create_vernacular_version` takes a collection ID, rather than a base version path, and constructs the base version path `<collection_id>/`. Since there is always only 1 data file per validation version, `create_validation_version` does not return a version path but instead a versioned csv name: `<mapped_metadata_version_path>/validation_<version suffix>.csv`

`rikolti.utils.versions` also implements a variety of functions to resolve a version path to an absolute URI for usage with the storage module, using environment variables to specify the particular rikolti data root: 

- `get_most_recent_vernacular_version(collection_id: Union[int, str])` sorts the contents of `$VERNACULAR_DATA/<collection_id>/`, and returns the version path of the first item - this presumes a sortable vernacular version suffix. Currently, datetime is used as the version suffix. 
- `get_pages(version: str, pipeline_step: str)` resolves a "vernacular" pipeline step version path to `$VERNACULAR_DATA/<version path>/` or a "mapped" pipeline step version path to `$MAPPED_DATA/<version path>/` and returns a list of version pages. `get_vernacular_pages` and `get_mapped_pages` just make sure the pipeline step is set correctly. 
- `get_child_directories(version, **kwargs)` resolves a mapped version path to `$MAPPED_DATA/<version path>/data/` and returns a list of directories in a given version's `data` folder. This is really only used to check if there is a "children" folder inside "data", so it could actually return a boolean and may be could be renamed, but I don't know what to call the resulting function. I anticipate having to revisit how versioning handles child records a bit when we revisit the Nuxeo fetcher/mapper pair (which is the only data source we support complex objects for). 
- `get_child_pages` resolves a version path to `$MAPPED_DATA/<version path>/data/children/` and returns a list of version pages. 
- `get_vernacular_page(version_page)` resolves a vernacular version page to `$VERNACULAR_DATA/<version page>` and returns the contents of that page
- `get_mapped_page(version_page)` resolves a mapped version page to `$MAPPED_DATA/<version page>` and returns the json-loaded contents of that page
- `put_page(content: str, page_name: Union[int, str], version: str, pipeline_step: str)`: resolves a version path to a page URI: `<data_root>/<version path>/data/<page_name>` and puts `contents` at that URI, where <data_root> is determined by `pipeline_step` to either $VERNACULAR_DATA, $MAPPED_DATA, or $CONTENT_DATA/
- `put_validation_report(content, version_page)`: resolves a version page to a URI: `$MAPPED_DATA/<data_root>/<version page>` and puts `contents` at that URI (there is no differentiation between a validation report's version path and a validation report's version page, since there is only ever one page per validation version). 

Questions:
- "pipeline step" strikes me as a bit of a weak abstraction within the `versions` module. I can imagine turning it into an Enum to make it a bit of a stronger abstraction, or de-drying the code a bit and just having separate `list`, `put`, and `get` functions for the different pipeline steps (see: https://github.com/ucldc/rikolti/pull/628/commits/4ed1e52876d2e8cbabf8e7f601b4be75e8656e09)
- I'm handling validation reports in a slightly unique way, I think this is OK and makes sense, but want to double check with you all. 
- version suffixes: we had discussed using dag run id, but this actually stays static even while we re-run task attempts. we could use dag run id and task attempt number. but if we wind up running this outside of the airflow context but writing to the s3 bucket, then sorting and determining the "most recent version" for example, is sorting over multiple version suffix namespaces. this PR currently just uses datetime as version suffix, but does provide a hook to provide a specified suffix. 
- partial-pipeline DAGs and getting the "most recent version" (or other sensible version?) vs. requiring our users to know and specify as a parameter a version path. In exactly once case here, I've implemented the "get most recent vernacular version" function which tries to understand the most recent vernacular version to pass into the partial-pipeline mapper DAG if no version path is supplied. This seems kind of bad, though, like there's lots of ways we could make some wrong assumptions about which version is most "sensible" to assume. Making our users understand how to get the version path for a partial-pipeline DAG simultaneously seems like a pretty big ask without really making version paths a regularly labeled and discussed part of the system. 

Some side-effects of this PR:
- removed `FETCHER_DATA_DEST`, `MAPPER_DATA_SRC`, `MAPPER_DATA_DEST`, `CONTENT_DATA_SRC`, `CONTENT_DATA_DEST`, and `CONTENT_DEST` environment variables
- added `VERNACULAR_DATA`, `MAPPED_DATA`, `CONTENT_DATA`, `CONTENT_ROOT` environment variables (where `CONTENT_ROOT` is the root for content files, while `CONTENT_DATA` is the root for metadata record files). 
- breaks the ability to run `python -m <component>/<module>.py`, because we can't import above the top level to get the utilities module. 
- required updates to the entrypoints and return values throughout the system in order to pass a particular vernacular data version as input to the mapper. previously the system just "knew" where to find vernacular data, so no parameter was necessary. even if we were still able to run from the command line, we would have to pass specific version paths as input to the command. 
- `metadata_mapper.UCLDCWriter` class is obsolete - it handled differences between writing to local vs. writing to s3 within the context of the metadata mapper in a specific way. Now that code has been generalized into the version system for mapper specific context and the storage system. 
- Much of the `metadata_mapper.Vernacular` class is obsolete - it handled differences between reading from local vs. reading from s3 within the context of the metadata mapper. It also handles parsing the read file, though, so the parsing part is still relevant. 
- Several of the `tests.py` files became challenging to contend with, due to the fact that they had certain expectations for where source files lived. These tests are outdated and not properly instrumented tests, though. In some cases I removed them entirely. 
- Most of Tim's `metadata_mapper.utilities` module is obsolete - it handled differences between writing and reading from local vs. s3 within the context of the validator. 
- `validate_registry_collections.py` module became challenging to contend with, but it hadn't been updated since my prototyped validator from ages ago - it had never even been updated to work with Tim's validator, so I just removed it. 
- had to add "create version" tasks prior to each dynamic pipeline step within the DAGs - for a pipeline step that functions for an entire collection, "create version" can happen within the python code prior to running the pipeline step. For a pipeline step that functions for a single page within a collection, "create version" has to happen /before/ the pipeline step so that we don't create a separate version (with a unique timestamp) for each page. We could maybe get around this if we used the dag run ID as the version suffix, but then we probably also want to use the task attempt number as well so that we can create new versions for each task attempt. 

Next Steps:
- as soon as this is put into place, we'll suddenly have a *ton* of data. I'm brainstorming ideas on how to moderate that. I think a "pin version" dag might be in order? It would somehow "pin" a specific version path (and all "data" folders down the version path) while deleting all the other versions. I can imagine multiple versions one might want to "pin" for a given collection (published version, version on -test, a known bad version that we're working to improve against), so I'm not sure if "pinning" is persistent and part of the versioning system (we move the directories into a different, "pinned" location? we suffix or prefix the directories with some special keyword?) or maybe the "pin version" dag takes a list of versions and just deletes everything else, and it's up to the user to keep track of what they want to keep and what they want to delete. 
- a "download version path" DAG that takes a given version path and zips up that version path while ignoring sibling version paths. 

Whew. Thanks for making it this far. Here's a cute hamster: https://www.instagram.com/reel/CzORAdxpk1o/?igshid=MzRlODBiNWFlZA==


utils/storage.py
```diff
@@ -0,0 +1,229 @@
...
+
+
+def put_file_content(data: DataStorage, content) -> str:
+    """
+    Write content to a file at data.path
+    """
+    file_path = os.sep.join(data.path.split('/'))
+    directory_path = os.path.dirname(file_path)
+    if not os.path.exists(directory_path):
+        os.makedirs(directory_path)
```
barbarahui, line 193: This needs to be:

`os.makedirs(directory_path, exist_ok=True)`

to avoid getting a FileExistsError when one of the intermediate dirs already exists, which happens when writing the 2nd page of data.

utils/versions.py
```diff
@@ -0,0 +1,156 @@
...
+        page_list = storage.list_pages(data_path, recursive=True, **kwargs)
+    except FileNotFoundError as e:
+        print(
+            f"\n\nNo {pipeline_step} pages found in {e.filename}\n\n"
+        )
+        raise(e)
+    return [path[len(data_root)+1:] for path in page_list]
+
+def get_vernacular_pages(version, **kwargs):
+    return get_pages(version, 'vernacular_metadata', **kwargs)
```
barbarahui, line 97: Should be:

`return get_pages(version, 'vernacular', **kwargs)`

utils/versions.py
```diff
@@ -0,0 +1,156 @@
...
+
+def get_mapped_pages(version, **kwargs):
+    return get_pages(version, 'mapped_metadata', **kwargs)
```
barbarahui, line 100: Should be: 

`return get_pages(version, 'mapped', **kwargs)`

barbarahui (commented):
OK, I've read through all of the code and tried running the `harvest_collection` dag and `map_collection` dag locally. The approach makes sense and looks good to me overall. This review includes various minor comments on the code itself. I'll leave some other more general comments separately.

utils/versions.py
```diff
@@ -0,0 +1,156 @@
...
+    data_path = f"{data_root.rstrip('/')}/{version.rstrip('/')}/data/children/"
+    try:
+        page_list = storage.list_pages(data_path, recursive=False, **kwargs)
+    except FileNotFoundError:
+        return []
+    except OSError:
+        return []
+    return [path[len(data_root)+1:] for path in page_list]
+
+def get_vernacular_page(version_page):
```
barbarahui, line 121: I think `get_vernacular_page_content` would be clearer as a name, and distinguish this function's purpose from `get_vernacular_pages`. 

You could also rename `get_vernacular_pages` --> `get_vernacular_page_list`? I think it's ok to leave as is too.

utils/versions.py
```diff
@@ -0,0 +1,156 @@
...
+    data_root = os.environ.get("VERNACULAR_DATA", "file:///tmp").rstrip('/')
+    return storage.get_page_content(f"{data_root.rstrip('/')}/{version_page}")
+
+def get_mapped_page(version_page):
```
barbarahui, line 125: Similarly, I think `get_mapped_page_content` would be clearer as a function name.

README.md
```diff
@@ -197,7 +193,7 @@ If you would like to run the content harvester on AWS infrastructure using the E
...
 
-If you would like to run your own rikolti/content_harvester image instead of pulling the image from AWS, then from inside the Rikolti repo, run `docker build -t rikolti/content_harvester content_harvester` to build the `rikolti/content_harvester` image locally and update the `content_harvester_image` to be `rikolti/content_harvester`.
+If you would like to run your own rikolti/content_harvester image instead of pulling the image from AWS, then from inside the Rikolti repo, run `docker build -f Dockerfile.content_harvester -t rikolti/content_harvester .` to build the `rikolti/content_harvester` image locally and update the `content_harvester_image` to be `rikolti/content_harvester`.
```
barbarahui, line 196: Small clarification:

If you would like to run your own rikolti/content_harvester image instead of pulling the image from AWS, then from inside the Rikolti repo, run `docker build -f Dockerfile.content_harvester -t rikolti/content_harvester .` to build the `rikolti/content_harvester` image locally and also add the following line to `dags/startup.sh`:

`export CONTENT_HARVEST_IMAGE=rikolti/content_harvester`

barbarahui:
Comments on your questions above:

**"pipeline step" strikes me as a bit of a weak abstraction**
Hrm yeah, there's already the bug I ran into with `get_vernacular_pages` using "vernacular_metadata" while get_pages uses "vernacular" as the pipeline step. I'm OK with both of your proposed changes: either (1) turning pipeline_step into an enum or (2) just having separate list, put and get functions for the different pipeline steps. My instinct is that the latter would be simpler to use and maybe more readable?

**I'm handling validation reports in a slightly unique way, I think this is OK and makes sense, but want to double check with you all.**
Seems fine to me!

**version suffixes**
On this topic, I've been thinking about index names. If we name the index after the corresponding content version, it'd be something unwieldy like this:

`27305/vernacular_metadata_2023-11-17T22:11:50/mapped_metadata_2023-11-17T22:12:00/content_data_2023-11-17T22:12:42`

Actually it would need to be something like this), because capital letters and forward slashes are disallowed:

`27305::vernacular_metadata_2023-11-17T22:11:50::mapped_metadata_2023-11-17-22:12:00::content_data_2023-11-17-22:12:42`

I'm in the process of trying to think of an elegant solution, but maybe a brainstorming session with you, me and @bibliotechy is in order.

**partial-pipeline DAGs and getting the "most recent version" (or other sensible version?) vs. requiring our users to know and specify as a parameter a version path**

Yeah, I see what you mean and don't have an elegant solution off the top of my head right now either. Am thinking on it...

utils/storage.py
```diff
@@ -0,0 +1,229 @@
...
+    """
+    List all objects in s3_bucket with prefix s3_prefix
+    """
+    s3 = boto3.client('s3', **kwargs)
+
+    s3_objects = s3.list_objects_v2(
+        Bucket=data.bucket, 
+        Prefix=data.path
+    )
+    # TODO: check resp['IsTruncated'] and use ContinuationToken if needed
```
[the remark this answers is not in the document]
amywieliczka: Made an issue: https://github.com/ucldc/rikolti/issues/644, unless you think this should be addressed as part of this PR.

utils/storage.py
```diff
@@ -0,0 +1,229 @@
...
+
+
+def put_page_content(content:str, data_uri: str, **kwargs) -> str:
+    """
+    Write content to a file at relative_path (relative to data_path).
+    relative_path is a list of strings, each string is a directory name 
+    representing a directory tree.
+    handle s3 or file storage, use '/' as separator for s3 key and os.sep
+    as separtors for file storage
+    """
```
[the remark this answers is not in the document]
amywieliczka: thanks, removed

metadata_fetcher/fetchers/Fetcher.py
```diff
@@ -25,58 +21,32 @@ class FetchError(Exception):
...
+            collection_id: str or int
+            write_page: str or int filename of the page to write to
+            vernacular_version: path relative to collection id
+                ex: "3433/vernacular_version_1"
+        """
         self.harvest_type = params.get('harvest_type')
         self.collection_id = params.get('collection_id')
         self.write_page = params.get('write_page', 0)
-        bucket = settings.DATA_DEST["BUCKET"]
+        self.vernacular_version = params['vernacular_version']
```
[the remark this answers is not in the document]
amywieliczka: Hmm, to your point about a nice error for a missing vernacular version, the only place where I see the fetcher class getting instantiated is here: https://github.com/ucldc/rikolti/blob/rikolti_versioning/metadata_fetcher/lambda_function.py#L42. The vernacular version is programmatically created a few lines up, rather than the result of user input. If there's some sort of failure in the creation of the vernacular version, then this _should_ result in developer intervention, imo. Unlike with the collection id, there's no place for the person running the fetcher to go provide the vernacular version. 

I could add the capacity for a user to specify a vernacular version, I suppose, but that sounds like a whole new feature request with a bunch of added complexity regarding how a harvesting operator might supply a version name.

utils/versions.py
```diff
@@ -0,0 +1,156 @@
+import json
+import os
+from datetime import datetime
+from typing import Union, Optional
+from . import storage
+
+def get_version(collection_id: Union[int, str], uri: str) -> str:
```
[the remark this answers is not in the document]
amywieliczka: Sure, since I wasn't entirely convinced I wouldn't wind up reworking the whole versioning system, I hadn't yet consistently incorporated comments. I'll address docstrings throughout momentarily.

metadata_fetcher/lambda_function.py
```diff
@@ -20,39 +21,48 @@ def import_fetcher(harvest_type):
 
 
 # AWS Lambda entry point
-def fetch_collection(payload, context):
+def fetch_collection(payload, vernacular_version, context) -> list[dict]:
```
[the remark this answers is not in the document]
amywieliczka: any good ideas? I do appreciate knowing that `lambda_` files generally contain entry points into the system.

content_harvester/content_types.py
```diff
@@ -0,0 +1,122 @@
...
+    def processed(self):
+        return (
+            self.derivative_filepath and 
+            os.path.exists(self.derivative_filepath)
+        )
+
+    def set_s3_filepath(self, s3_filepath):
+        self.s3_filepath = s3_filepath
+
+    def __bool__(self):
```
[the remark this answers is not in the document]
amywieliczka: Used here: https://github.com/ucldc/rikolti/blob/rikolti_versioning/content_harvester/by_page.py#L72

content_harvester/content_types.py
```diff
@@ -0,0 +1,122 @@
...
+        return not self.missing
+
+    def __del__(self):
```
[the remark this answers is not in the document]
amywieliczka: I believe this is also used in `by_page`. I only moved these classes out of `by_page.py` and refactored out storage concerns in this PR, I did not re-work the classes themselves.

utils/storage.py
amywieliczka: Strange, I don't get that error. Will do.

bibliotechy:
> **partial-pipeline DAGs and getting the "most recent version" (or other sensible version?) vs. requiring our users to know and specify as a parameter a version path**

I feel like the use case for the partial pipeline dags is for testing and troubleshooting fetching or mapping or indexing without having to recreate all of the previous steps. Is that a correct assumption? Or do we expect the partial dags to be used for actual getting to data to production processes? 

If we expect to use them only for testing, then I think it is reasonable to expect the use to provide a path. Maybe, if a path is not provided, the task can get a list of recent paths and suggest options?

utils/versions.py
```diff
@@ -75,29 +61,17 @@ def get_most_recent_vernacular_version(collection_id: Union[int, str]):
...
 
-def get_pages(version: str, pipeline_step: str, **kwargs):
-    if pipeline_step == 'vernacular': 
-        data_root = os.environ.get("VERNACULAR_DATA", "file:///tmp")
-    elif pipeline_step == 'mapped':
-        data_root = os.environ.get("MAPPED_DATA", "file:///tmp")
-    else:
-        raise Exception("Invalid pipeline step")
-
+def get_vernacular_page(version, **kwargs):
```
barbarahui, line 64: I think this should be `get_vernacular_pages`?

barbarahui (commented):
I think there's just one typo in a function name in `versions.py` (see comment). Otherwise this looks good to me!

utils/versions.py
amywieliczka: Hey @barbarahui - are you familiar with suggestions? You can make one line changes like this as a suggestion in your PR response. It's an option to the left of H B I (heading, bold, italics) in this comment box.

amywieliczka:
```suggestion
def get_vernacular_pages(version, **kwargs):
```

barbarahui: Ah! Will def use suggestions in future!

barbarahui (approved):

Issue #581, opened 2023-10-26
Generate and review validation reports: burbank_islandora


christinklez:
- **Step 2: Generate reports** [status: in progress --> ready for review]
  - [x] Generate validation reports (run `validate_by_mapper DAG`): https://7a8067cb-3b99-477e-a883-7e311175a9b4.c3.us-west-2.airflow.amazonaws.com/dags/validate_by_mapper_type/grid?dag_run_id=manual__2023-11-13T22%3A25%3A28%2B00%3A00
  - [x] Add a comment (in this issue) to the output URLs
  - [x] Download the validation reports locally; prepend the collection registry number to the filename
  - [x] Upload validation reports to the [Rikolti Data Validation](https://drive.google.com/drive/folders/1AgDPsz6oyDhsfgRuLyjomaXYgE1sPLNT) folder
- **Step 3: Review reports**  [status: ready for review --> review in progress]
  - [x] Review [sample collections validation reports](https://docs.google.com/spreadsheets/d/1XkBwi8jiuGgrWvQBqjkrbavSl13yyrQNyRyEb_CQ4z0/edit#gid=1143782640); leave notes in the spreadsheet
  - [x] Quickly (and not very deeply) scan through other validation reports (to identify if any new metadata fields surface)
  - [x] Synthesize notes at the [mapper level](https://docs.google.com/spreadsheets/d/1XkBwi8jiuGgrWvQBqjkrbavSl13yyrQNyRyEb_CQ4z0/edit#gid=1100474517)
  - [x] Share an update (and surface any anomalies) during a weekly check-in (Mondays at 3:00pm)
- **Step 4: Have data validation issues? Address & resolve validation issues (synthesized mapper-level):**
  - [ ] Create a new issue (“Validation results: Rikolti mapper name“) and report/request CIC fixes & investigations. Assign to CIC.
  - [ ] Add a comment to link to the new “validation results” issue.
  - [ ] Add a “blocked/paused” label to this issue. Include a status note in this issue title (e,g., HOLD for CIC fixes; HOLD for code review; etc.) 
  - [ ] Move this issue back to the “To Do” pile.
- [Re-generate & re-review (repeat steps 2-3 above, as needed.]
- **Step 5: Have acceptable validation reports?**
  - [x] Move the valid (accepted) reports to the “VALID” folder in google drive.
[...]

PR #649, opened 2023-11-28
[bugfix] bubank_mapper: collate_fields callable not called

barbarahui:


barbarahui:
Fixes https://github.com/ucldc/rikolti/issues/581#issuecomment-1809254642

PR #674, opened 2023-12-17
Implement MARC fetcher and mappers

lthurston:
Includes:

- MARC fetcher (for old-fashioned machine-readable, MARC)
- Mapper for ucb_tind, at the time of writing in a very early state

Notes:

- The code for the legacy ucb_tind mapper is nearly as convoluted as MARC is archaic. This represents an attempt to come at the mapper from a more developer-friendly perspective (a goal, as I understand it, of Rikolti generally). However, it also means that the rewrite may take some time to make feature complete.
- Class structure is not final. Specifically, ucb_tind, should be a subclass of OAI, maybe? And, development is happening in the ucb_tind mapper, but some of the word might apply to the marc parent class. This work is more a POC for the approach to the mapping.

### DIFF
--- a/dags/requirements.txt
+++ b/dags/requirements.txt
@@ -4,6 +4,10 @@ opensearch-py
 requests
 sickle
 python-dotenv
+beautifulsoup4
+lxml
+pymarc
+MarkupSafe
 apache-airflow-providers-docker
 apache-airflow-providers-google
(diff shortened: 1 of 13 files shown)


amywieliczka:
Closing per @barbarahui's comments in #1142

PR #684, opened 2024-01-03
Feat: Non-blocking errors

amywieliczka:
- Report out errored collections but continue harvesting
- Update documentation

### DIFF
--- a/dags/utils_by_mapper_type.py
+++ b/dags/utils_by_mapper_type.py
@@ -45,16 +45,32 @@ def fetch_endpoint_task(endpoint, params=None):
     limit = params.get('limit', None) if params else None
     fetcher_job_result = fetch_endpoint(endpoint, limit, logger)
     fetched_versions = {}
-    for collection_id in fetcher_job_result.keys():
+    errored_collections = {}
+    for collection_id, fetch_job in fetcher_job_result.items():
+        if isinstance(fetch_job, dict) and 'error' in fetch_job:
+            errored_collections[collection_id] = fetch_job
+            continue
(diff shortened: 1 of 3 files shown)


dags/utils_by_mapper_type.py
```diff
@@ -45,16 +45,32 @@ def fetch_endpoint_task(endpoint, params=None):
...
-    for collection_id in fetcher_job_result.keys():
+    errored_collections = {}
+    for collection_id, fetch_job in fetcher_job_result.items():
+        if isinstance(fetch_job, dict) and 'error' in fetch_job:
+            errored_collections[collection_id] = fetch_job
+            continue
+        if not isinstance(fetch_job, list):
+            errored_collections[collection_id] = fetch_job
+            continue
+        if not fetch_job[0].get('vernacular_filepath'):
```
bibliotechy, line 56: Is there any possible case where `fetch_job` is an empty list that you should guard against here?

bibliotechy (approved):
One question, but looks good

dags/utils_by_mapper_type.py
amywieliczka: It's possible that `fetch_job[0]` is an empty list, but not that `fetch_job` is empty. I don't know what it would mean for `fetch_job` to be an empty list - it is a list of fetch page results. If it were to be empty, then that would mean the fetcher hadn't run for a single page? At which point, I think a blocking error is appropriate. I'm not even sure what a "helpful" error message might be in that case, because I can't imagine why it would happen. It probably means the Fetcher is declaring that it's finished before it's even started?

PR #685, opened 2024-01-03
Set response encoding to utf-8 for OAI feeds

amywieliczka:
Allow aggregate_vernacular_content to manipulate response encoding


metadata_fetcher/fetchers/oai_fetcher.py
```diff
@@ -62,6 +62,7 @@ def build_fetch_request(self):
             url += f"&metadataPrefix={self.metadata_prefix}"
             if self.metadata_set:
                 url += f"&set={self.metadata_set}"
+            print(url)
```
bibliotechy, line 65: Do you want to keep this print?

bibliotechy (approved):
One question about a print statement. Otherwise looks good

PR #703, opened 2024-01-19
Barbara/publishing pipeline

barbarahui:
I believe the code is in place to start moving staged collection indices to the `rikolti_prd` alias! To do this, run the `index_collection_to_prod` DAG in Airflow.

The 2 changes introduced with this PR are: 

1. using AWS credentials for auth when talking to the OpenSearch API
2. assigning `id` rather than `calisphere-id` as the `_id` when creating OpenSearch records

Otherwise, the functionality to promote an index to production was already in place.

### DIFF
--- a/record_indexer/README.md
+++ b/record_indexer/README.md
@@ -22,6 +22,8 @@ Add the current stage index for a collection to the `rikolti-prd` alias:
 python -m record_indexer.move_index_to_prod <collection_id>
 ```
 
+Note that the index creation code enforces the existence of one stage index at a time. This means we can simply supply the collection ID as input and the process will move the current stage index to production.
+
 
 
 
--- a/record_indexer/README.md
+++ b/record_indexer/README.md
@@ -8,7 +8,7 @@ python index_templates/rikolti_template.py
 
 This creates a template that will be used whenever an index with name matching `rikolti*` is added to the cluster.
 
-## Run indexer
+## Run indexer from command line
 
 Create a new index for a collection and add it to the `rikolti-stg` alias:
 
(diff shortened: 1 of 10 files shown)


record_indexer/README.md
```diff
@@ -22,6 +22,8 @@ Add the current stage index for a collection to the `rikolti-prd` alias:
 python -m record_indexer.move_index_to_prod <collection_id>
 ```
 
+Note that the index creation code enforces the existence of one stage index at a time. This means we can simply supply the collection ID as input and the process will move the current stage index to production.
```
amywieliczka, line 25: Could you elaborate on the lifecycle of an index? I know we've discussed all this but I can't quite remember the rules. 

We create a new index with a stage alias each time we run the harvesting pipeline, presumably also deleting any existing indices with a stage alias? 

At some point we're happy with a stage index and we want to push it to production, so we run the publishing dag, which adds the prod alias to the index, and removes the stage alias? so now we have no indices with a stage alias but one index with a prod alias? 

At some point we repeat this process, and when we add the prod alias to the new index that's ready for publishing, then the index with the prod alias becomes a plain unaliased index? these unaliased indices are kept around per the retention policy sorted by date? so as a rule, any index without an alias is a previously published index? and, as a rule, if that index retention policy were set to more than 1, we could sort any index without an alias by date to, for example, restore the most recently published index?

amywieliczka (approved):
Just a request for documentation, really, otherwise looks good.

record_indexer/README.md
barbarahui: README is updated.

PR #714, opened 2024-01-27
Calisphere etl

barbarahui:
The mapper is ready for review.

I see what you mean about the solr API returning a final empty page using nextCursorMark. This causes problems when running the `harvest_collection` DAG where the mapper fans out by page. I can't figure out what the heck is going on with Solr.

Since the solr index is static and this is transitional functionality that we'll retire after we cutover to rikolti, I went ahead and added a bit of a hacky workaround to the fetcher's `increment()`  method, which reports "finished" if we've fetched the number of docs that we're expecting. (See this commit: https://github.com/ucldc/rikolti/pull/714/commits/2fe62d02604a53130e98b3e51f21ac3c01d0adc5).

### DIFF
--- a/metadata_mapper/lambda_function.py
+++ b/metadata_mapper/lambda_function.py
@@ -116,7 +116,8 @@ def map_page(
 
     # TODO: analyze and simplify this straight port of the
     # solr updater module into the Rikolti framework
-    mapped_records = [record.solr_updater() for record in mapped_records]
+    if collection.get('rikolti_mapper_type') != 'calisphere_solr.calisphere_solr':
+        mapped_records = [record.solr_updater() for record in mapped_records]
     mapped_records = [record.remove_none_values() for record in mapped_records]
 
     group_page_exceptions = {}
--- a/metadata_mapper/lambda_function.py
+++ b/metadata_mapper/lambda_function.py
@@ -56,7 +56,9 @@ def parse_enrichment_url(enrichment_url):
 
 
 def run_enrichments(records, collection, enrichment_set, page_filename):
-    for enrichment_url in collection.get(enrichment_set, []):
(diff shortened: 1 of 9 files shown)


metadata_mapper/lambda_function.py
```diff
@@ -56,7 +56,8 @@ def parse_enrichment_url(enrichment_url):
 
 
 def run_enrichments(records, collection, enrichment_set, page_filename):
-    for enrichment_url in collection.get(enrichment_set, []):
+    enrichment_urls = collection.get(enrichment_set) or []
```
bibliotechy, line 59: You've done this in several places, replacing `collection.get(some_value, [])` with `collection.get(some_value) or []`. Is there a difference in behavior between those two? It isn't obvious to me what that is.  No issue with it, just wondering.

bibliotechy (approved):
One question. Looks good to me.

amywieliczka (approved):
This looks good to me, but similarly confused by the replacement of `<dict>.get(<key>, [])` with `<dict>.get(<key>) or []`

metadata_mapper/lambda_function.py
barbarahui: @amywieliczka @bibliotechy so yeah, it turns out that if you explicitly set the value to `None` then this is what happens:

```
Python 3.9.0 (default, May 23 2023, 15:28:21)
[Clang 13.0.0 (clang-1300.0.27.3)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> mydict = {"foo": None}
>>> for x in mydict.get("foo", []):
...    print(x)
...
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: 'NoneType' object is not iterable
```

Whereas:
```
>>> mydict.get("foo") or []
[]
```

PR #723, opened 2024-01-31
S3 pagination

barbarahui:
This PR adds pagination to the `s3.list_objects_v2()` in `utils.storage.list_s3_pages()` so that we now get results over 1000.

It also fixes a bug that I ran into in `utils.versions.get_version()` where splitting on `f"{collection_id}/` would sometimes return a list of more than 2 elements, e.g. for the following uri:

```8/vernacular_metadata_2024-01-31T00:39:58/data/986```

### DIFF
--- a/utils/storage.py
+++ b/utils/storage.py
@@ -98,20 +98,23 @@ def list_s3_pages(data: DataStorage, recursive: bool=True, **kwargs) -> list:
     List all objects in s3_bucket with prefix s3_prefix
     """
     s3 = boto3.client('s3', **kwargs)
-    s3_objects = s3.list_objects_v2(
+    paginator = s3.get_paginator('list_objects_v2')
+    pages = paginator.paginate(
         Bucket=data.bucket, 
         Prefix=data.path.lstrip('/')
     )
-    # TODO: check resp['IsTruncated'] and use ContinuationToken if needed
 
-    keys = [f"s3://{data.bucket}/{obj['Key']}" for obj in s3_objects['Contents']]
-    prefix = f"s3://{data.bucket}/{data.path}"
+    keys = []
+    for page in pages:
+        keys += [f"s3://{data.bucket}/{obj['Key']}" for obj in page['Contents']]
 
     if not recursive:
         # prune deeper branches
+        prefix = f"s3://{data.bucket}/{data.path}"
         leaf_regex = re.escape(prefix) + r"^\/?[\w!'_.*()-]+\/?$"
         keys = [key for key in keys if re.match(leaf_regex, key)]
 
+    print(f"len(keys): {len(keys)}")
     return keys
 
 
--- a/utils/storage.py
+++ b/utils/storage.py
@@ -105,7 +105,7 @@ def list_s3_pages(data: DataStorage, recursive: bool=True, **kwargs) -> list:
     )
 
     keys = []
-    for page in pages:
+    for obj_list in page_iterator:
         keys += [f"s3://{data.bucket}/{obj['Key']}" for obj in page['Contents']]
 
     if not recursive:
--- a/utils/storage.py
+++ b/utils/storage.py
@@ -106,7 +106,7 @@ def list_s3_pages(data: DataStorage, recursive: bool=True, **kwargs) -> list:
 
     keys = []
     for obj_list in page_iterator:
-        keys += [f"s3://{data.bucket}/{obj['Key']}" for obj in page['Contents']]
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: utils/versions.py [hunk 1]
```diff
@@ -27,15 +27,12 @@ def get_version(collection_id: Union[int, str], uri: str) -> str:
...
     """
     collection_id = str(collection_id)
-    uri = uri.rstrip('/')
-    uri_parts = uri.lstrip('/').split('/')
+    uri_parts = uri.strip('/').split('/')
     if str(collection_id) not in uri_parts or len(uri_parts) < 2:
         raise Exception(f"Not a valid version path: {uri}, {uri_parts}")
-    rikolti_data_root, relative_path = uri.split(f"{collection_id}/")
-    path_list = relative_path.split('/')
+    path_list = uri_parts[uri_parts.index(collection_id):]
```
amywieliczka, line 33:
```suggestion
    path_list = uri_parts[uri_parts.index(f"{collection_id}/"):]
```

Okay, so this screwed me over at one point: I had to include the `/` because sometimes collection 233 has a `233/vernacular_v1/data/233.json` page number. 

How you have it implemented here should be fine, because `index` should snag the _first_ instance of `collection_id`, but the suggestion makes it a bit more explicit.

amywieliczka: Oh, actually, I think this isn't quite right, either. `data` needs to be a discrete path part, not just in the path somewhere. `233/vernacular_v1/mapped_metadata_v1/` includes `data`.

amywieliczka: So we actually do want to split on `/` and then check if `data` is in the path list.

amywieliczka: Was there something broken with this implementation of `get_version`? It was pretty tricky to implement, so if it's not broken, I don't think it's worth fixing.

amywieliczka (changes requested):

utils/versions.py
```diff
@@ -27,15 +27,12 @@ def get_version(collection_id: Union[int, str], uri: str) -> str:
     Returns a version path.
     """
     collection_id = str(collection_id)
-    uri = uri.rstrip('/')
-    uri_parts = uri.lstrip('/').split('/')
+    uri_parts = uri.strip('/').split('/')
     if str(collection_id) not in uri_parts or len(uri_parts) < 2:
         raise Exception(f"Not a valid version path: {uri}, {uri_parts}")
-    rikolti_data_root, relative_path = uri.split(f"{collection_id}/")
-    path_list = relative_path.split('/')
```
amywieliczka, line 35: This isn't quite right, either. `data` needs to be a discrete path part, not just _in_ the path somewhere. `data in 233/vernacular_metadata_v1/mapped_metadata_v1/ == True`. Truncating to `233/vernacular_meta` would be bad. 

So we actually do want to split on `/` and then check if `data` is in the path list. 

Was there something broken with this implementation of `get_version`? It was pretty tricky to implement, so if it's not broken, I don't think it's worth fixing.

amywieliczka (changes requested):

utils/storage.py
```diff
@@ -98,17 +98,19 @@ def list_s3_pages(data: DataStorage, recursive: bool=True, **kwargs) -> list:
     List all objects in s3_bucket with prefix s3_prefix
     """
     s3 = boto3.client('s3', **kwargs)
-    s3_objects = s3.list_objects_v2(
+    paginator = s3.get_paginator('list_objects_v2')
+    pages = paginator.paginate(
```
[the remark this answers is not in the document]
amywieliczka: Thanks so much for doing this! I found `page_iterator` to be a bit more clear than `pages` and `obj_list` to be a bit clearer than `page`, but it's gravy.

utils/versions.py
barbarahui: Hrm, the way I have it implemented, `f"{collection_id}/` isn't going to be in `path_list` because we're splitting on `/`.

`path_list` is going to look something like:
```
["8", "vernacular_metadata_2024-01-31T00:39:58", "data", "986"]
```

So we just need to find the index of `{collection_id}`

utils/versions.py
```diff
@@ -27,15 +27,12 @@ def get_version(collection_id: Union[int, str], uri: str) -> str:
     Returns a version path.
```
amywieliczka, line 27:
```suggestion
    Returns a version path.
    Edge cases: "8/vernacular_metadata_2024-01-31T00:39:58/data/986", "8/vernacular_metadata_v1/data/8"
```

PR #749, opened 2024-02-01
Handle empty md pages

barbarahui:
This PR updates the codebase to allow for vernacular pages that do not contain any metadata records. In this case:
- the fetcher issues a warning and writes the API reponse to storage. I figure that storing the page will be helpful for troubleshooting and verifying whether or not the the empty page is harmless
- the mapper issues a warning and mapping/enrichment is skipped
- a mapped_metadata page is not written to storage and `mapped_page_path` in the return dict is set to `None`

The code is also updated in several places to handle a None value for `mapped_page_path`, where before the code was always expecting a string value.

Finally, it fixes a of bug in `mapper.py` where exceptions in enrichments were not getting raised, which was obscuring a problem in `replace_substring()`.

### DIFF
--- a/dags/shared_tasks.py
+++ b/dags/shared_tasks.py
@@ -152,7 +152,7 @@ def map_page_task(vernacular_page: str, collection: dict, mapped_data_version: s
         status: success
         num_records_mapped: int
         page_exceptions: TODO
-        mapped_page_path: str, ex: 
+        mapped_page_path|None: str, ex:
             3433/vernacular_metadata_2023-01-01T00:00:00/mapped_metadata_2023-01-01T00:00:00/1.jsonl
     """
     collection_id = collection.get('id')
@@ -170,7 +170,7 @@ def get_mapping_status_task(collection: dict, mapped_pages: list):
         status: success
         num_records_mapped: int
         page_exceptions: TODO
-        mapped_page_path: str, ex: 
+        mapped_page_path: str|None, ex:
             3433/vernacular_metadata_2023-01-01T00:00:00/mapped_metadata_2023-01-01T00:00:00/1.jsonl
     returns a dict with the following keys:
         mapped_page_paths: ex: [
--- a/dags/shared_tasks.py
+++ b/dags/shared_tasks.py
@@ -233,8 +233,9 @@ def validate_collection_task(collection_id: int, mapped_metadata_pages: dict) ->
 
 @task()
 def create_with_content_urls_version_task(collection: dict, mapped_pages: list[dict]):
-    mapped_version = get_version(
-        collection['id'], mapped_pages[0]['mapped_page_path'])
+    mapped_page_path = [page['mapped_page_path'] for page in mapped_pages
+        if page['mapped_page_path']][0]
+    mapped_version = get_version(collection['id'], mapped_page_path)
     return create_with_content_urls_version(mapped_version)
 
 
--- a/metadata_fetcher/fetchers/Fetcher.py
+++ b/metadata_fetcher/fetchers/Fetcher.py
@@ -60,15 +60,20 @@ class Fetcher(object):
                 f"[{self.collection_id}]: unable to fetch page {page}")
 
         record_count = self.check_page(response)
+        if not record_count:
+            logger.warning(
+                f"[{self.collection_id}]: no records found "
+                f"on page {self.write_page}"
+            )
+
         filepath = None
-        if record_count:
-            content = self.aggregate_vernacular_content(response)
-            try:
-                filepath = put_vernacular_page(
-                    content, self.write_page, self.vernacular_version)
-            except Exception as e:
-                print(f"Metadata Fetcher: {e}")
-                raise(e)
+        content = self.aggregate_vernacular_content(response)
+        try:
+            filepath = put_vernacular_page(
+                content, self.write_page, self.vernacular_version)
+        except Exception as e:
+            print(f"Metadata Fetcher: {e}")
+            raise(e)
 
         self.increment(response)
 
--- a/metadata_mapper/lambda_function.py
+++ b/metadata_mapper/lambda_function.py
@@ -116,52 +116,52 @@ def map_page(
             'page_exceptions': {},
             'mapped_page_path': None,
         }
-    else:
-        source_metadata_records = run_enrichments(
-            source_metadata_records, collection, 'rikolti__pre_mapping', page_filename)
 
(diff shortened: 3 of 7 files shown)


dags/shared_tasks.py
```diff
@@ -170,7 +170,7 @@ def get_mapping_status_task(collection: dict, mapped_pages: list):
         status: success
         num_records_mapped: int
         page_exceptions: TODO
-        mapped_page_path: str, ex: 
+        mapped_page_path: str|None, ex:
```
amywieliczka, line 173: I get errors when I do this, but you can `from typing import Optional` and then `mapped_page_path: Optional[str]`

metadata_fetcher/fetchers/Fetcher.py
```diff
@@ -60,15 +60,20 @@ def fetch_page(self):
                 f"[{self.collection_id}]: unable to fetch page {page}")
 
         record_count = self.check_page(response)
+        if not record_count:
+            logger.warning(
+                f"[{self.collection_id}]: no records found "
+                f"on page {self.write_page}"
+            )
```
amywieliczka, line 67: Does it makes sense to raise an error instead? 

`raise ValueError(f"[{self.collection_id}]: no records found on page {self.write_page} at {response.request.url}")`

amywieliczka: Also, I don't see the logger messages in Airflow, do you??? I only see `print` messages.

metadata_mapper/lambda_function.py
```diff
@@ -105,50 +105,63 @@ def map_page(
...
-        'status': 'success',
-        'num_records_mapped': len(mapped_records),
-        'page_exceptions': group_page_exceptions,
-        'mapped_page_path': mapped_page_path,
-    }
+    if not source_metadata_records:
+        logging.warning(
+            f"No source vernacular records found for {collection_id} "
+            f"page {vernacular_page_path}."
+            )
```
amywieliczka, line 112: Same thing re: seeing log messages in Airflow log UI.

metadata_mapper/lambda_function.py
```diff
@@ -105,50 +105,63 @@ def map_page(
...
+        return {
+            'status': 'success',
+            'num_records_mapped': 0,
+            'page_exceptions': {},
+            'mapped_page_path': None,
+        }
+    else:
```
amywieliczka, line 119: Due to the return statement inside the `if` block above, you can get rid of this `else` and save on the tabs.

metadata_mapper/mappers/mapper.py
```diff
@@ -142,9 +142,10 @@ def enrich(self, enrichment_function_name: str, **kwargs):
         func: Callable = getattr(self, enrichment_function_name)
         try:
             return func(**kwargs)
-        except Exception as e:
-            print(f"ENRICHMENT ERROR: {str(kwargs)}")
-            raise e
+        except Exception:
```
amywieliczka, line 145:
```suggestion
        except Exception as e:
```

metadata_mapper/mappers/mapper.py
```diff
@@ -142,9 +142,10 @@ def enrich(self, enrichment_function_name: str, **kwargs):
...
+            raise Exception(f"ENRICHMENT ERROR for enrichment: `{enrichment_function_name}` "
```
amywieliczka, line 146:
```suggestion
            raise e(f"ENRICHMENT ERROR for enrichment: `{enrichment_function_name}` "
```

metadata_mapper/mappers/mapper.py
```diff
@@ -142,9 +142,10 @@ def enrich(self, enrichment_function_name: str, **kwargs):
...
+                f"with kwargs: `{str(kwargs)}` "
+                )
```
amywieliczka, line 148: This is why I want https://docs.python.org/3/library/exceptions.html#BaseException.add_note and https://docs.python.org/3/library/exceptions.html#ExceptionGroup

amywieliczka: I think this means we do lose the exception that was thrown by the enrichments, though? Like if it was a ValueError or a KeyError or whatnot.

amywieliczka (dismissed):
You seem to have veered towards logging, printing, and returning none instead of raising exceptions. 

I do kinda think that if `fetch_page` can't fetch a page, it should raise an exception? And then it's for the code that invokes `fetch_page` to choose what it wants to do with that exception (print an error and continue on with whatever else it's doing or halt). I've had developing opinions on this, and can't say I've sat down to think it all through holistically, and I think the codebase (or at least the parts I've written) reflects some of this ongoing development of opinions in it's mashup of different tactics. 

Anyway, aside from this difference of opinion, the only other comment I have is on logging, which I've not managed to get quite right with Airflow logs. Perhaps I have something misconfigured, but I only seem to see `print()` statements in the airflow logs. I'm curious to check if the log statements you've added in this PR come through for me.

metadata_fetcher/fetchers/Fetcher.py
barbarahui: Yes, I see logger messages here. I only couldn't get logger messages to appear in `metadata_mapper/validate_mapping.py`, which is why that's a print statement.

metadata_mapper/lambda_function.py
barbarahui: Well, but the whole point is that it's perfectly valid for some pages to be empty after parsing. There are several collections where all of the records on a particular page are deleted, so the parser skips/removes them.

barbarahui: But true, if we configured mapping to continue even if an error is thrown, this would be flagged more visibly.

metadata_mapper/mappers/mapper.py
barbarahui: No, the exception thrown by enrichments does appear in the logs. I tested it out.

barbarahui:
@amywieliczka Hmm yeah, I see what you mean about raising an exception if the metadata portion of the page is empty and then having the invoking code deciding whether or not to continue. I was operating on the fact that pages with no mappable metadata (i.e. pages where all of the records are marked as deleted) are sometimes perfectly valid. But you're right, if an empty page is invalid then it'd be easy to overlook if you don't notice the warning messages.

barbarahui:
@amywieliczka I created this card for us to revisit MVP: https://github.com/ucldc/rikolti/issues/753

PR #751, opened 2024-02-02
Update OAI fetcher to throw error if no records found

amywieliczka:
@christinklez this should address the Quartex issue we were just talking about in standup

### DIFF
--- a/metadata_fetcher/fetchers/Fetcher.py
+++ b/metadata_fetcher/fetchers/Fetcher.py
@@ -52,6 +52,9 @@ class Fetcher(object):
             f"[{self.collection_id}]: fetching page {self.write_page} "
             f"at {page.get('url')}"
         )
+        if 'url' not in page:
+            raise InvalidHarvestEndpoint(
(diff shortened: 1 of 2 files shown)


amywieliczka:
@barbarahui this is a bit of a duplicate of your empty metadata page handling. I'm going to let that land first since it's a bit more encompassing and then update this after. `OaiFetcher.check_page` could return 0 instead of raising an error.

barbarahui:
@amywieliczka This is compatible with the work I did re. empty pages. The change I made to the Fetcher in that PR was to write the vernacular page to S3 if we were able to fetch one, regardless of the number of records that check_page() finds. I also added a warning to the log if the number of records is 0, where previously we had neither a warning nor an error.

Since in this PR you raise an error in the oai_fetcher rather than Fetcher, this works just fine with what's in main for now. We should still go back at some point and do a holistic review of our approach and be consistent about how we handle pages without metadata (both at the fetching and mapping stage).

barbarahui (approved):

PR #756, opened 2024-02-02
Update is_shown_by for omeka mapper

barbarahui:


barbarahui:
Bypassed review for this one.

Issue #197, opened 2022-11-08
Migrate enrich_date and enrich_earliest_date enrichments


amywieliczka:
Should be migrated as methods on the Record class in `metadata_mapper/mapper.py`

Related to https://github.com/ucldc/rikolti/pull/196

amywieliczka:
I performed some analysis of the usage of `enrich_earliest_date` and `enrich_date` enrichment functions, and then migrated these enrichments from calisphere-legacy-harvester/dpla-ingestion as part of this PR: https://github.com/ucldc/rikolti/pull/1090

Next steps:
- This is simply a theoretical migration of the enrichments, and no harvests have actually been run through these migrated enrichment functions. We need to test-run several collections. 
- Get the validator working again against Solr (beanstalk solr is still alive and kicking) and Couch (may need to start up `ingest-front-prd`) so that we can validate the facet_decade, sort_date_start, and sort_date_end fields for records - maybe even get this working in deployed mwaa, so @aturner and @christinklez can validate these fields? 
- Determine either a) a migration path to update records already in the index or b) a re-harvesting plan to progressively pick up decade data for harvests on re-harvest of the collections. 
- Update the Calisphere UI to facet and filter on the `facet_decade` field instead of the `date` field ([Stage ucldc/public_interface#405](https://github.com/ucldc/public_interface/pull/405))
- [No further work needed] Confirmed that Calisphere UI already uses `sort_date_start` and `sort_date_end` to support the sort options - no changes needed for this, though as more records have these fields, we'll get better sorts.

amywieliczka:
Collection Date Samples - issue #1093 - https://docs.google.com/spreadsheets/d/1JNw7ynxCJS8d4f5W4Glwwqx0qMm8ss-aB_GpjVxN1dI/edit?gid=0#gid=0

amywieliczka:
Completed since last update:

- Got the code actually working and running in my local mwaa runner (fixed my local mwaa runner)
- Got the validator working again
  - Made validator conditional on the presence of UCLDC_SOLR_URL and UCLDC_COUCH_URL
  - Commented out `is_shown_at` and `is_shown_by` validations (these validated against couch, which is no longer up)
- Ran the validator for [27 test collections](https://docs.google.com/spreadsheets/d/1JNw7ynxCJS8d4f5W4Glwwqx0qMm8ss-aB_GpjVxN1dI/edit?gid=0#gid=0) - filter on Calisphere Solr Mapper Type = No

Next steps:

[...]

PR #768, opened 2024-02-17
Validation updates

amywieliczka:
I ran validation reports with the updated validations @barbarahui added and there were only two issues:

1) `sort_date_start` and `sort_date_end` were reliably `None` in `rikolti`, which is expected, and confirms the work identified here: https://github.com/ucldc/rikolti/issues/197, which I believe we've determined is Post-MVP. These are used in the "Sort By" dropdown in determining Oldest First or Newest First. I've commented them out in the validation definition, and added a TODO. 

2) Qualified Collection and Repository IDs: 

| Solr Value | Rikolti Value |
| ---- | ---- | 
| `https://registry.cdlib.org/api/v1/collection/3433/` | `3433` |
| `https://registry.cdlib.org/api/v1/repository/4/::Repository Name::Campus Name` | `4::Repository Name::Campus Name` |
| `Sortable Collection Name:Collection Display Name:https://registry.cdlib.org/api/v1/collection/6/` | `Sortable Collection Name:Collection Display Name:6` |

I've made the decision to use an unqualified collection and repository identifier in Rikolti, and this PR, with it's custom validator function, codifies that. 

I believe it is safe to remove this qualified identifier because:
- The Calisphere UI never uses this fully qualified identifier - it always strips out the URL and just uses the numerical ID
- Unqualified Collection IDs are used through Rikolti to mean "the collection with this ID in registry-prd", so this assumption is already bound up in Rikolti throughout the codebase. 
- These identifiers are quite stable, and to the extent that we have a responsibility to maintain collection landing pages at stable URLs (ex: https://calisphere.org/collections/28011/), they will remain stable. 
- If this kind of qualification becomes necessary in the future (for example, maybe we want to harvest from registry-stg differently than we harvest from registry-prd and we want the resultant records to know that they were harvested using collection harvesting information from registry-stg instead of from registry-prd), we can add an additional field to each record, something like: `"registry-location": "https://registry.cdlib.org/api/v1/"`

This will require some minor updates in the Calisphere UI (which currently strips out the URL part of the string in order to get at the numerical part), but overall, I believe it will reduce a lot of complexity in Calisphere's UI, which is continually translating registry URLs into numerical IDs and back for the sake of interacting with Solr.


barbarahui (approved):
This makes sense to me! Looks good.

bibliotechy:
@amywieliczka Considering the granularity of indexing collections, it is very easy to eliminate a collection mistakenly harvested from stg and return to a previous index, so I think this is fine if it simplifies other things.

bibliotechy (approved):

PR #771, opened 2024-02-26
Delete a partially created index in exception handler, not airflow

amywieliczka:


record_indexer/settings.py
```diff
@@ -13,5 +17,18 @@ def get_auth():
     return AWSV4SignerAuth(
         credentials, os.environ.get("AWS_REGION", "us-west-2"))
 
+
+def print_opensearch_error(r: requests.Response, url: str):
```
bibliotechy, line 21: Not a blocker, but this feel weird in a `settings.py`

bibliotechy (approved):

record_indexer/settings.py
amywieliczka: Agreed, I wasn't sure where it did belong. I suppose I could create a `record_indexer.utils`

PR #793, opened 2024-03-07
Add thumbnail dimensions; add pillow req to content_harvester img

amywieliczka:
Doesn't add `reference_image_dimensions` to the Open Search index, but does add `dimensions` into the `thumbnail` datastructure, which I believe goes directly into Open Search?


barbarahui (approved):
Yup, that should work! Thanks!

PR #848, opened 2024-03-27
Capture PIL errors and print calisphere-id

barbarahui:
@amywieliczka I wanted to run this by you before merging since I know you've done some work on how best to approach capturing exceptions and bubbling up the info to the airflow logs. I think it's pretty straightforward though? Printing the calisphere-id along with the error from PIL will let us hone in on which specific image is problematic.

### DIFF
--- a/content_harvester/by_record.py
+++ b/content_harvester/by_record.py
@@ -3,6 +3,7 @@ import os
 from typing import Optional
 
 from PIL import Image
+from PIL import UnidentifiedImageError
 import requests
 from requests.adapters import HTTPAdapter, Retry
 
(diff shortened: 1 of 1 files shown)


amywieliczka (commented):
How you've implemented error handling looks just fine to me! I think I would have /maybe/ put it a level up since all these instances are part of `harvest_record_content` - so in `by_page.py`:

```
for i, record in enumerate(records):
    try:
        record_with_content = harvest_record_content(record, collection_id, rikolti_mapper_type)
    except UnidentifiedImageError as e:
        raise Exception(
            f"PIL.UnidentifiedImageError for calisphere-id "
            f"{record['calisphere-id']}: {e}"
        )
    except Image.DecompressionBombError as e:
        raise Exception(
            f"PIL.Image.DecompressionBombError for calisphere-id "
            f"{record['calisphere-id']}: {e}"
        )
```
This is also where we would want to make record-level decisions about how to proceed with harvesting content for this page if this error is thrown - for example, maybe instead of raising an exception, we print an error message and `continue` in the for loop to the next record in the page. 

An argument for keeping the exception handling logic where you have it is to include some context about which case is failing: an image thumbnail, a generated pdf thumbnail, or a generated video thumbnail - but I don't know that this extra contextual information about the three different failure cases is actually all that relevant.

amywieliczka (approved):

PR #914, opened 2024-05-03
Add migrate-stage-indexes script

barbarahui:


bibliotechy (approved):
Only request would be some comments about the expected env_vars with some example values.

PR #918, opened 2024-05-07
Create single multi collection stg index

barbarahui:
This PR does a few things:
- Adds `record_indexer/scripts/combine_individual_collection_stage_indices.py`, which combines the existing individual collection indices into one combined index. We'll probably only ever run this once.
- Adds `record_indexer/update_stage_index.py`, which updates this single index. It assumes we are using an alias and that only one index is associated with it at a time
- Makes `STAGE_ALIAS` a setting. This is so I could add a temporary `STAGE_ALIAS_COMBINED` setting to use while we are transitioning.
- Adds an indexer task that calls the new `record_indexer/update_stage_index.py` code.

If/when we decide to cut over to this single index, we'll need to:

1. Halt harvesting
2. Create a new combined stage index by running `record_indexer/scripts/combine_individual_collection_stage_indices.py`
3. Unalias the old individual collection indices from `rikolti-stg`
4. Alias the new combined index to `rikolti-stg`
5. Resume harvesting
6. Delete the old individual collection indices

I did create a combined stage index yesterday to test out the script. It is named `rikolti-stg-combined-20240506134027` and aliased to `rikolti-stg-combined`. I tried pointing my local instance of Calisphere at this alias by setting `ES_ALIAS=rikolti-stg-combined` in my local env, and it was happy. Queries seem a bit snappier too, but that's anecdotal.


record_indexer/scripts/combine_individual_collection_stage_indices.py
```diff
@@ -0,0 +1,46 @@
...
+            "dest":{
+                "index": combined_index
+            }
+        }
+        url = f"{endpoint}/_reindex"
+        headers = {"Content-Type": "application/json"}
+
+        r = requests.post(url, headers=headers, data=json.dumps(data), auth=auth)
+        print(f"{r.json()}\n")
+        r.raise_for_status()
```
amywieliczka, line 38: I recall opensearch provides really thorough details regarding why something has failed in the response text - I think instead, we want to add to the imports at the top: `from .utils import print_opensearch_error`, and then here:

```
    r = requests.post(url, headers=headers, data=json.dumps(data), auth=auth)
    if not (200 <= r.status_code <= 299):
        print_opensearch_error(r, url)
        r.raise_for_status()
```

record_indexer/scripts/combine_individual_collection_stage_indices.py
```diff
@@ -0,0 +1,46 @@
...
+import json
+import os
+import sys
+
+import requests
+
+def main():
+    ''' combine all indices aliased to `rikolti-stg` into one large index '''
+    endpoint = os.environ.get("OPENSEARCH_ENDPOINT")
+    auth = (os.environ.get("OPENSEARCH_USER"), os.environ.get("OPENSEARCH_PASS"))
```
amywieliczka, line 11: I have, in my env.local

`RIKOLTI_ES_ENDPOINT`, `RIKOLTI_ES_USER`, and `RIKOLTI_ES_PASS`, I also see a `settings.get_auth()`

What are these `OPENSEARCH_` values supposed to be?

record_indexer/update_stage_index.py
```diff
@@ -0,0 +1,48 @@
...
+
+    # add pages of records to index
+    for version_page in version_pages:
+        add_page(version_page, index)
+
+def get_index_for_alias(alias: str):
+    # for now, there should be only one index per alias (stage, prod)
+    url = f"{settings.ENDPOINT}/_alias/{alias}"
+    r = requests.get(url, auth=settings.get_auth())
+    r.raise_for_status()
```
amywieliczka, line 23: might want to add `print_opensearch_error` here.

amywieliczka (changes requested):
I think you could probably just update these lines in the `create_stage_index_task` https://github.com/ucldc/rikolti/blob/02f813faa0461b8b72322a0a8ca059e7a00e5d61/dags/shared_tasks/indexing_tasks.py#L23-L28 to read:

```py
try:
    update_stage_index_for_collection(collection_id, version_pages)
except Exception as e:
    # TODO: implement some rollback exception handling?
    raise e
```

then you wouldn't have to make any changes to `harvest_dag.py`. 

If you're really set on creating a new `_task` function to match the `update_stage_index` function, then please do copy the log output and registry communication pieces from the `create_stage_index_task` function. 

These lines provide some handy log output for searching our opensearch index: https://github.com/ucldc/rikolti/blob/02f813faa0461b8b72322a0a8ca059e7a00e5d61/dags/shared_tasks/indexing_tasks.py#L31-L41

And this line tells the Collection Registry that the collection has been indexed successfully: https://github.com/ucldc/rikolti/blob/02f813faa0461b8b72322a0a8ca059e7a00e5d61/dags/shared_tasks/indexing_tasks.py#L43 - the second argument could simply be `{status: 'success'}` if you no longer think it makes sense to return the index name.

barbarahui:
I addressed Amy's review requests. Chad took a look and says all looks good. I'm going to go ahead and merge this into main without waiting for re-review from Amy since she's out sick.

PR #943, opened 2024-05-14
Parse out prefix from filepaths to shorten payload sent to ECS

barbarahui:


amywieliczka (approved):
Ah! So neat and tidy!

PR #954, opened 2024-05-22
948 update sort collection data

amywieliczka:
Update sort_collection_data for ETL'ed collections from 

`<sortable name>::<display name>::https:://registry.cdlib.org/api/v1/collection/<collection id>`

to 

`<sortable name>::<display name>::<collection id>`

and revert any instances of `::` in the display name back to `:`.

### DIFF
--- a/content_harvester/by_registry_endpoint.py
+++ b/content_harvester/by_registry_endpoint.py
@@ -4,19 +4,7 @@ import requests
 
 from .by_collection import harvest_collection_content
 from rikolti.utils.versions import get_most_recent_mapped_version
-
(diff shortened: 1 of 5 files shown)


amywieliczka:
Hey @barbarahui - the vast majority of this has already been run and done, but I did want you to take a look at the `registry_client` module I added! 

`registry_client.registry_endpoint` is a bad name for that function, but it's a small refactor. I think maybe ultimately we want something like a `Registry` (or `RegistryClient`) class with a `.collections()` method?

barbarahui (approved):

PR #1116, opened 2024-09-30
Relax content cache: require ETag OR Last-Modified, not both

amywieliczka:


amywieliczka:
Bypassing branch protections - this is a small change @bibliotechy and I already discussed on Slack.

PR #1117, opened 2024-10-01
OS Schema is nested under index name

amywieliczka:
Bugfix for create_stage_index task


amywieliczka:
Bypassing codeowner review requirement to get bugfix in.

PR #1123, opened 2024-10-02
Don't use default value when doing dict lookup for 'thumbnail_source'

barbarahui:
I am still scratching my head over why this is the case, but it seems that if you use a function call as the default value for `dict.get()`, then the function call always runs, even if the key exists in the dictionary.

[Update]: OK so this makes sense. This is because, for any function call, python evaluates the arguments before the function call is executed. So for the following statement:

```
thumbnail_src = record.get('thumbnail_source', get_thumb_src(record))
```

Python runs `get_thumb_src(record)` first, before running `record.get()`.

Running `get_thumb_src(record)` updates the `record` dict "in place", so that the value of `record['thumbnail_source']` becomes:

```
{'url': {'url': 'https://nuxeo.cdlib.org/nuxeo/nxfile/default/19c41b7c-36ef-470d-bced-761046a3eb72/file:content/mss86-4_005_021.pdf?changeToken=55-0', 'mimetype': 'application/pdf', 'filename': 'mss86-4_005_021.pdf', 'nuxeo_type': 'CustomFile'}}
```
The code downstream then doesn't know how to handle a url value of type dict.

An alternative solution would be to make a deepcopy of the `record` dict:

```
import copy
thumbnail_src = record.get('thumbnail_source', get_thumb_src(copy.deepcopy(record)))
```

### DIFF
--- a/content_harvester/by_record.py
+++ b/content_harvester/by_record.py
@@ -84,7 +84,6 @@ def harvest_record_content(
             collection_id,
             rikolti_mapper_type: Optional[str] = None,
             ) -> dict:
-
     # Weird how we have to use username/pass to hit this endpoint
     # but we have to use auth token to hit API endpoint
     auth = None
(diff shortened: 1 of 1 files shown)


amywieliczka:
Ohh, yup that makes sense. 

I definitely like the `if not...` construction rather than your `deepcopy` alternative. No point wasting resources on a deepcopy. 

Makes me wonder the details on how python evaluates functions in the case of the ternary operator: 

`thumbnail_src = record.get('thumbnail_source') or get_thumb_src(record)`

amywieliczka (approved):

amywieliczka:
Seems that in the ternary operator, the or'ed condition is not evaluated until the prior condition is False. 

```python
def test():
    print('inside test')

test_val = True or test()
test_val = False or test()
# inside test
```

PR #1131, opened 2024-10-11
Strip date string after splitting on ;

barbarahui:
Date values like this: ["1494; 1495"] look like this after enrichment: ["1494", " 1495"]

Note the space before "1495". Not sure how critical it is to remove this, but the validation report was flagging it as a problem, and maybe it affects the decade faceting?


amywieliczka (approved):
looks good to me! thanks for looking at the validation reports.

PR #1135, opened 2024-10-15
Delete tmpfiles

barbarahui:


content_harvester/by_record.py
```diff
@@ -106,13 +106,15 @@ def harvest_record_content(
...
         request = ContentRequest(thumbnail_src_url, auth)
-        record['thumbnail'] = create_thumbnail_component(
+        record['thumbnail'], tmp_files = create_thumbnail_component(
             collection_id,
             request,
             thumbnail_src,
             record['calisphere-id']
         )
 
+    [os.remove(filepath) for filepath in tmp_files]
```
amywieliczka, line 116: So one problem with this route is that `tmp_files` could be undefined in the case where there's no media and no thumbnail. 

Another problem is that, if there is a media file and a thumbnail, any tmp_files returned by `create_media_component` will be overwritten by the tmp_files returned by `create_thumbnail_component`, and won't be deleted here at line 116.

amywieliczka (changes requested):

content_harvester/by_record.py
barbarahui: D'oh, of course. How does this look?

amywieliczka: Looks good, but now, in the case where the media_source and the thumbnail_source are the same, you might wind up with duplicates in `tmp_files` and `os.remove` will throw a `FileNotFoundError` when it tries to delete the same file a second time. 

`[os.remove(filepath) for filepath in set(tmp_files)]` would de-dupe the list, or you could explicitly check `if os.path.exists(filepath)` before removing it.

amywieliczka (changes requested):

content_harvester/by_record.py
barbarahui: Egads, right you are! Pushed another commit.

amywieliczka (approved):

PR #1139, opened 2024-10-18
Aggressive caching for Nuxeo only

amywieliczka:
Alternative to PR #1138 but for Nuxeo only

### DIFF
--- a/content_harvester/by_record.py
+++ b/content_harvester/by_record.py
@@ -1,6 +1,6 @@
 import hashlib
 import os
-from datetime import datetime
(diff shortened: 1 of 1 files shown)


barbarahui:
This PR should resolve the Nuxeo caching issue: https://github.com/ucldc/rikolti/pull/1140

amywieliczka:
I'm assuming from @christinklez's Slack message this past Monday that we do want recency based caching for all content, not just Nuxeo content, which makes this PR moot (and doubly so if etags are now fixed for Nuxeo)

PR #1140, opened 2024-10-21
ETag checks: allow redirects and add Nuxeo workaround

barbarahui:
By default `requests` will perform location redirection for all verbs except HEAD, so I've added `allow_redirects=True` to the etag check request.

Also, it turns out that S3 presigned URLs for GET requests are different from those for HEAD requests. The Nuxeo API redirects us to a presigned S3 URL, which returns a 403 because it's only good for GETs. The easiest thing I could figure out to get around this was to issue a GET request that only returns the headers (basically mimicking the behavior of a HEAD request).


content_harvester/by_record.py
```diff
@@ -221,7 +221,15 @@ def check_component_cache(
...
-            head_resp = http_session.head(**asdict(request))
+            if 'nuxeo' in request.url:
+                # The S3 presigned URLs from Nuxeo are good for GET requests only
+                # so do a GET request that mimics a head request
+                head_resp = http_session.get(
+                    **asdict(request),
+                    headers={"Range": "bytes=0-0"}
+                )
+            else:
+                head_resp = http_session.head(**asdict(request), allow_redirects=True)
```
bibliotechy, line 232: requests doc indicate that is `True` by default 

> allow_redirects ([bool](https://docs.python.org/3/library/functions.html#bool)) – (optional) Boolean. Enable/disable GET/OPTIONS/POST/PUT/PATCH/DELETE/HEAD redirection. Defaults to True.

Why wasn't that working before?

bibliotechy (approved):

content_harvester/by_record.py
barbarahui: In the docs I’m looking at (maybe not the version we’re on, but) it says “ By default Requests will perform location redirection for all verbs except HEAD.”

https://docs.python-requests.org/en/latest/user/quickstart/#redirection-and-history

bibliotechy: Lol, it's the same site - https://docs.python-requests.org/en/latest/api/#module-requests

PR #1149, opened 2024-11-04
Remove couchdb env vars

barbarahui:
This PR removes use of `UCLDC_COUCH_URL` and `UCLDC_COUCH_TIMEOUT` env vars from the codebase, as CouchDB is no longer running.

When running the `harvest_collection` dag, the validation report generation was getting skipped. We decided to reinstate it.

When running the validator, the reports will be created without validating `isShownAt` and `isShownBy`, since those validations rely on CouchDB data. 

I did not actually do the work of stripping out all of the logic relating to couchdb. I figure it makes sense to do that when we overhaul the validator to compare against OpenSearch data.

There is also a 2nd commit in this PR that I think fixes a logic error: https://github.com/ucldc/rikolti/pull/1149/commits/64277b55ef80e2fc83a5711e9adba214d5206e20

### DIFF
--- a/dags/shared_tasks/mapping_tasks.py
+++ b/dags/shared_tasks/mapping_tasks.py
@@ -265,7 +265,7 @@ def validate_endpoint_task(url, mapped_versions, params=None, **context):
         print(f"please validate manually: {list(errored_collections.keys())}")
         print("*" * 60)
 
-    if not len(errored_collections) == len(validations):
+    if len(errored_collections) == len(validations):
         print("-", file=sys.stderr)
         raise ValueError("No collections successfully validated, exiting.")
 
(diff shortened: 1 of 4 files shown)


dags/shared_tasks/mapping_tasks.py
```diff
@@ -266,7 +265,7 @@ def validate_endpoint_task(url, mapped_versions, params=None, **context):
         print(f"please validate manually: {list(errored_collections.keys())}")
         print("*" * 60)
 
-    if not len(errored_collections) == len(validations):
+    if len(errored_collections) == len(validations):
```
barbarahui, line 268: @amywieliczka It seems like this was an error in logic, but I might not be understanding what exactly this is trying to achieve?

amywieliczka (approved):

PR #1153, opened 2024-11-12
Fetcher retries

barbarahui:
- implement retries in base Fetcher class
- move `make_http_request()` functionality into `ucd_json_fetcher` -- this is the only place this was being used

### DIFF
--- a/metadata_fetcher/fetchers/Fetcher.py
+++ b/metadata_fetcher/fetchers/Fetcher.py
@@ -44,10 +44,21 @@ class Fetcher(object):
         self.collection_id = params.get('collection_id')
         self.write_page = params.get('write_page', 0)
         self.vernacular_version = params['vernacular_version']
-
+        self.http_session = self.configure_http_session()
 
         if not self.collection_id:
             raise CollectionIdRequired("collection_id is required")
+
+    def configure_http_session(self) -> requests.Session:
+        http = requests.Session()
+        retry_strategy = Retry(
+            total=3,
+            status_forcelist=[413, 429, 500, 502, 503, 504],
+        )
+        adapter = HTTPAdapter(max_retries=retry_strategy)
+        http.mount("https://", adapter)
+        http.mount("http://", adapter)
+        return http
 
     def fetch_page(self) -> FetchedPageStatus:
         """
@@ -68,11 +79,13 @@ class Fetcher(object):
             f"at {page.get('url')}"
         )
         try:
-            response = requests.get(**page)
+            response = self.http_session.get(**page)
             response.raise_for_status()
-        except requests.exceptions.HTTPError:
+        except self.http_session.exceptions.HTTPError as e:
             raise FetchError(
-                f"[{self.collection_id}]: unable to fetch page {page}")
+                f"[{self.collection_id}]: unable to fetch page {page}",
+                f"Error was: {e}"
+            )
 
         record_count = self.check_page(response)
         if not record_count:
--- a/metadata_fetcher/fetchers/Fetcher.py
+++ b/metadata_fetcher/fetchers/Fetcher.py
@@ -4,9 +4,8 @@ import requests
 from dataclasses import dataclass
 from typing import Optional
 
-from requests.adapters import HTTPAdapter, Retry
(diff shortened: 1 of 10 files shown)


metadata_fetcher/fetchers/Fetcher.py
```diff
@@ -44,11 +44,22 @@ def __init__(self, params: dict):
...
+        self.http_session = self.configure_http_session()
 
         if not self.collection_id:
             raise CollectionIdRequired("collection_id is required")
 
+    def configure_http_session(self) -> requests.Session:
+        http = requests.Session()
+        retry_strategy = Retry(
+            total=3,
+            status_forcelist=[413, 429, 500, 502, 503, 504],
```
amywieliczka, line 56: Why did you remove the back off factor - that seems useful and relevant?

metadata_fetcher/fetchers/ucd_json_fetcher.py
```diff
@@ -89,7 +90,10 @@ def fetch_json_ld(self, url: str) -> Optional[dict]:
         Returns:
             dict
         """
-        response = self.make_http_request(url)
+        session = requests.Session()
+        retries = Retry(total=3, backoff_factor=2)
+        session.mount("https://", HTTPAdapter(max_retries=retries))
+        response = session.get(url=url)
```
amywieliczka, line 96: If you added the backoff_factor to the gloabl configuration for `self.http_session`, then you should be able to do: 

```suggestion
        response = self.http_session.get(url=url)
```

And get rid of lines 93-95 here. 

Sure, `self.http_session` adds some status codes to the forcelist and attached the retry strategy to both http and https, but that seems fine?

metadata_fetcher/fetchers/Fetcher.py
amywieliczka: I also wonder if this makes sense as a global configuration - like in `rikolti/utils.py`?

amywieliczka (changes requested):

amywieliczka:
Looks like `nuxeo_fetcher`, `oac_fetcher`, and `ucd_json_fetcher` all make their own calls to `requests.get`, so you might want to update those other instances in `nuxeo_fetcher` and `oac_fetcher` to use `self.http_session.get` instead.

metadata_fetcher/fetchers/Fetcher.py
barbarahui: Well, I didn't actually remove the backoff factor as there wasn't one configured at all here previously. I figured that we could go with the default backoff factor of 0 and tweak if necessary. But sure, it probably makes sense to set it to `2`.

Yeah, we could make this a global configuration. This is actually a copy of `configure_http_session()` from the content harvester code (which doesn't have a backoff factor set).

metadata_fetcher/fetchers/ucd_json_fetcher.py
barbarahui: Yeah, I thought about that too. I figured to introduce the least amount of change possible, but you're right, it probably would be fine.

barbarahui:
@amywieliczka Sure yeah, I think it makes sense to do this. I can do the work to put this in place. I don't think it'll take too long--I was just trying to do this quickly so that I could focus on the Nuxeo API issue.

metadata_fetcher/fetchers/ucd_json_fetcher.py
amywieliczka: Mmmm, that makes sense - what you did change (`Fetcher.fetch_page()`) is the function that is called like 95% of the time though, so it seemed confusing to me to change it most of the way but not all of the way. Could result in weird errors on strange edge cases where we've had to implement additional requests outside of the standard requests managed by the Fetcher base class.

amywieliczka:
Figured I could do this quickly while @barbarahui's head was in Nuxeo API issues. 
Resolved my own asks - @barbarahui I'll leave this open in case you want to take a look, but as far as I'm concerned it's good to go - rebase & merge when you get to it.

barbarahui:
@amywieliczka this looks great, thank you so much!!!!

PR #1163, opened 2025-01-11
Update nuxeo fetcher to use db query lambda

barbarahui:
The main difference here, other than hitting the dbquery lambda endpoint, is that recursing down the hierarchy of documents is implemented here in the fetcher code rather than in the lambda.


bibliotechy (approved):

barbarahui:
@christinklez @aturner The nuxeo fetcher update has been deployed to Airflow. If you could test out harvesting a few Nuxeo collections, that would be great. Please let me know if you run into any problems.

PR #1214, opened 2026-01-10
Update validation csv reports to QA markdown reports

amywieliczka:
Replaces validation reports with 2 QA reports: a summary report and a details report. 

The QA reports are broken out into 3 different sections: 
- Item-Level Diff: compare `id` fields (the id used in Calisphere URLs) to determine dropped or changed IDs and new or changed IDs. Prints a list of both the `calisphere-id` (the source id, an OAI id we get from a provider, for example) and the `id` fields, but only compares `id` fields (since we care about keeping URLs as consistent as possible). 
- Field-Level Requirements: checks that all candidate records (the most recently harvested metadata in this job run) contain the required fields and prints a list of both the `calisphere-id` and `id` fields for any records missing required metadata fields. 
- Field-Level Diff: compare each record against the last published version of the record to determine what has been updated or changed (if anything)

The basis of comparison for the Item-Level Diff and the Field-Level Diff is either:
- the currently indexed metadata (the metadata record is retrieved directly from the index). In this case, we cannot compare `is_shown_at`, `is_shown_by`, `item_count`, `media_source`, or `thumbnail_source` (in the candidate metadata), and `rikolti`, `thumbnail`, or `media` (in the indexed metadata). 
- the mapped version of the metadata that is currently published in the index. 

The mapped version of the metadata that is currently published in the index is the ideal basis of comparison, but in many cases we do not know the mapped version of the metadata that is currently published in the index, since the published version in the index is simply identified as "initial", rather than a full s3 path. Therefore, the currently indexed metadata must be used, and the fields mentioned above must be scrubbed (since the thumbnail url in the mapped metadata is the contributing institution's thumbnail url, but the thumbnail url in the indexed metadata is _our_ s3 url, for example). 

Caveats & Future Work:
- Requires adding deepdiff to airflow's requirements.txt [**high priority - needs to happen before we merge this PR**]
- I've placed this diff report module directly in the `/dags/shared_tasks/` directory, but maybe it should go into the metadata_mapper directory like the validation reports? 
- The Field-Level Diff identifies and matches records for comparison by `calisphere-id`, rather than by `id`. This should probably preference `id` (since this is what we care about vis-a-vis persistent calisphere urls) but make `calisphere-id` an option (so we can still get this kind of diff reporting for cases where we accept an `id` change and a break in persistent urls). 
- While I've replaced the validation reports in the workflow, I have not stripped validation reports out of the codebase. At some point we should remove this code. 
- While I'm internally calling these "QA reports" (a combo of diffs and requirements-checking), they are still called "validation" in the DAG workflow. When we get DAG versioning in Airflow v3, we should rename this task. 
- Add markdown rendering to ucldc/aws-js-s3-explorer: #1216 
- DeepDiff acknowledges a few different kinds of diffs, but `values_changed` is the only one we aggregate on (any others are reported in the record-specific diff dump at the end of the report), because it's the only one that came up in the sample data. I anticipate continuing to iterate on the aggregation mechanism as other types of diffs come up.

### DIFF
--- a/dags/requirements.txt
+++ b/dags/requirements.txt
@@ -9,3 +9,4 @@ apache-airflow-providers-docker
 apache-airflow-providers-google
 apache-airflow-providers-amazon
 pymarc
+deepdiff


dags/shared_tasks/diffs.py
```diff
@@ -0,0 +1,687 @@
...
+
+def _get_indexed_data_from_opensearch(calisphere_ids: list[str]):
+    search_url = f"{settings.ENDPOINT}/rikolti-prd/_search"
+    query = {
+        "query": {
+            "terms": {
+                "calisphere-id": calisphere_ids
+            }
+        },
+        "size": len(calisphere_ids) * 2
```
barbarahui: I can't for the life of me figure out: why is the size set to 2 times the number of calisphere_ids here?

barbarahui (approved):
Looks good to me!! Just a random question.

dags/shared_tasks/diffs.py
amywieliczka: Hmmm, this may have been cruft from a previous iteration, and since it doesn't really matter, it never got removed. There was a time when I was trying to figure out different ways to test for id diffs and record diffs simultaneously (building a list of IDs as I went and also compared them), but ultimately I felt that the code was difficult to read and reason about and the complexity wasn't worth saving an opensearch query.

PR #1228, opened 2026-02-04
Add caltech mapper

barbarahui:


metadata_mapper/mappers/oai/islandora/caltech_mapper.py
```diff
@@ -0,0 +1,72 @@
...
+            fields = defaultdict(list)
+            for element in metadata_elements:
+                tag = re.sub(r'\{.*\}', '', element.tag)
+                if tag == 'identifier':
+                    id_type = element.get('type')
+                    if id_type == 'Web-Access':
+                        fields['isShownAt'].append(element.text)
+                    elif id_type == 'image-thumbnail':
+                        fields['isShownBy'].append(element.text)
+                fields[tag].append(element.text)
```
amywieliczka: At a philosophical level, I'm pretty opposed to doing this kind of mapping logic here in the parser - mapping should happen in the mapper. I understand that there is additional data in the element's type attribute, but maybe fields['identifier_type-Web-Access'], and fields['identifier_type-image-thumbnail']

This would keep the mapping logic in the mapper, and keeps the source language congruent with the source, even while changing the source data structure type.

PR #1229, opened 2026-02-05
Add archipelago mapper

barbarahui:


amywieliczka (changes requested):
Same comments re: mapping logic happening in the parser. Also, what is the difference between this one and the caltech one?

PR #1306, opened 2026-04-21
Youtube mapper updates

### LINKED ISSUE #1296: `youtube` mapper issues
Generally for all youtube collections
https://github.com/ucldc/rikolti/blob/main/metadata_mapper/mappers/youtube/youtube_mapper.py

## `isShownAt URL`
- Needs to use  `videoId: iQmUeBGYetE` instead of `id: UExHcHlJNzRObmVsNGhOOUEwWmJOTmp1emdGcW1OaFlmYi4zMDg5MkQ5MEVDMEM1NTg2` to construct the URL
- See example: 
  - stage: https://calisphere-stage.cdlib.org/item/7f0638f60bc5ce4e6428ba4436f4e504/ points to https://www.youtube.com/watch?v=UExHcHlJNzRObmVsNGhOOUEwWmJOTmp1emdGcW1OaFlmYi4zMDg5MkQ5MEVDMEM1NTg2 (doesn't work)
  - prod: https://calisphere.org/item/5a99db6f8f17391c705d65658d642154/ points to https://www.youtube.com/watch?v=iQmUeBGYetE (works)
- to do: update line 17 & 18 `id` to `videoId`

## `title`, `description` (possibly other fields, but haven't seen any)
- displays each character of a word in a different line
- see example https://calisphere-stage.cdlib.org/item/7f0638f60bc5ce4e6428ba4436f4e504/

## question about `identifier`
- the legacy mapper previously didn't map `identifier` - the re-harvest is now resulting in now Calisphere IDs + item record page URLs
- Question: Should we map `identifier`? If so, could it be the `videoId`?
- Maybe update line 8?

barbarahui:


metadata_mapper/mappers/youtube/youtube_mapper.py
```diff
@@ -4,33 +4,54 @@
...
         tags = self.source_metadata.get("snippet", {}).get("tags", [])
 
         return [{"name": tag} for tag in tags]
 
     def map_title(self):
-        return self.source_metadata.get("snippet", {}).get("title")
+        title = self.source_metadata.get("snippet", {}).get("title")
+
+        return self.string_to_list(title)
+
```
amywieliczka:
```suggestion

# TODO: make more robust and add to mapper.py as a general purpose data-transformation tool (with other general purpose data transformation tools such as collate_fields)
```

amywieliczka (dismissed):

metadata_mapper/mappers/youtube/youtube_mapper.py
amywieliczka: Oh, just looked at the next three updates, which all copy this code and duplicate it across 3 different mappers. Could we just add this to the base mapper class? 

Also, string to list is, I think, a dangerous misnomer here. It implies that it exclusively takes strings, and exclusively returns lists of strings. But actually, it takes any value at all, and returns that value unchanged, _unless_ it's a string (in which case it does return a list). I'm not actually sure what the intended behavior is here. On the one hand, you could change the function implementation to match the function name (throw an error if anything is passed that is not a string, change strings to lists, and return lists - in which case, I think I would just ditch this function altogether in favor of using wrapping `[]` on the metadata mapping), or you could change the function name to be something like "listify_any_string_values(values)" - but even that feels a little bit bad. 

I assume you can't just replace calls to `string_to_list(title)` with `[title]` because of the type ambiguity of the source metadata ie: sometimes `title` is already a list, sometimes it's just a string? 

What about: 
```python
from typing import Union

def ensure_list(value: Union[list, str]) -> list:
    if isinstance(value, list):
        return value
    elif isinstance(value, str):
        return [value]
    raise ValueError(f"Cannot ensure_list of {type(value)}: {value}")
```

If you want to be extra: `ensure_list_of_strings`

amywieliczka: I think a goal of mapping is to move from ambiguous data types to discrete data types - not punt ambiguity further along the pipeline. Our legacy harvester did a lot of punting ambiguity along, but this made it really difficult to trace where, exactly, issues occurred. Everything tried to handle every data type. 😓 😵‍💫

barbarahui: > I assume you can't just replace calls to string_to_list(title) with [title] because of the type ambiguity of the source metadata ie: sometimes title is already a list, sometimes it's just a string?

Yeah, exactly. I like your suggestion of adding the function to the base mapper class. Will do and resubmit the PRs for review!

amywieliczka (approved):

PR #1309, opened 2026-05-01
Add ensure_list() mapper helper function

barbarahui:


amywieliczka:
Hmmm, I would expect `ensure_list` to operate on the field values, not the field name. I think `ensure_field_is_list` is a better name for what you've implemented here (though I think this is less useful to your YT mapper case)

I think you can just abandon the @returns_callable construct and use `def ensure_list(field_value)` and do the same `spatial: self.ensure_list(self.source_metadata.get("place")` style mapping that you had been doing with `self.string_to_list` (is there something about CIC's mapper construct that prevents that?)

We don't have other fieldvalues-based general mapper functions in mapper.py, but I think we could. I just want to be very clear which are operating on values and which are operating on fields, and in general, these functions should always reduce ambiguity about data type.

barbarahui:
Yeah, I had just come to the same conclusion as you independently, without having seen your comment!

amywieliczka (approved):
🎉

PR #1323, opened 2026-05-18
Youtube updates

### LINKED ISSUE #1319: `youtube` mapper/fetcher issue & question
## missing thumbnails
**Request:**
The mapper grabs the `standard` thumbnail option, and maps to `is_shown_by`, see: https://github.com/ucldc/rikolti/blob/main/metadata_mapper/mappers/youtube/youtube_mapper.py#L30

Would it be possible to add a fallback option? If the `standard` option is unavailable, perhaps use `high` as the next best option?

**Back story:**
The vernacular record generally includes up to 5 thumbnail URL options, see this example from collection 27030:
```
        "thumbnails": {
          "default": {
            "url": "https://i.ytimg.com/vi/iGDuOmObPp0/default.jpg",
            "width": 120,
            "height": 90
          },
          "medium": {
            "url": "https://i.ytimg.com/vi/iGDuOmObPp0/mqdefault.jpg",
            "width": 320,
            "height": 180
          },
          "high": {
            "url": "https://i.ytimg.com/vi/iGDuOmObPp0/hqdefault.jpg",
            "width": 480,
            "height": 360
          },
          "standard": {
            "url": "https://i.ytimg.com/vi/iGDuOmObPp0/sddefault.jpg",
            "width": 640,
            "height": 480
          },
          "maxres": {
            "url": "https://i.ytimg.com/vi/iGDuOmObPp0/maxresdefault.jpg",
            "width": 1280,
            "height": 720
          }
        },
```

However, some vernacular records seem to only have up to 3 thumbnail URL options (from the same collection 27030:
```
        "thumbnails": {
          "default": {
            "url": "https://i.ytimg.com/vi/lZto3ZJryfo/default.jpg",
            "width": 120,
            "height": 90
          },
          "medium": {
            "url": "https://i.ytimg.com/vi/lZto3ZJryfo/mqdefault.jpg",
            "width": 320,
            "height": 180
          },
          "high": {
            "url": "https://i.ytimg.com/vi/lZto3ZJryfo/hqdefault.jpg",
            "width": 480,
            "height": 360
          }
        },
```

## lost `subject` - vernacular does not include `tags` data
**Back story:**
With the exception of 27022 (which is unique, because it's a single video that's not coming from a playlist) - it seems that that at least 14 collections have lost the subject field.

From 27022: The vernacular is providing `tags` fields, example here:
```
        "tags": [
          "sfpl",
          "san francisco public library",
          "san francisco",
          "sfpl.org",
          "Hunters Point (Neighborhood)",
          "Bayview (Neighborhood)"
        ],
```

The mapper maps `tags` to `subject`: https://github.com/ucldc/rikolti/blob/main/metadata_mapper/mappers/youtube/youtube_mapper.py#L40

However, the other collections do not have the `tags` field in the fetched vernacular data. It also does not seem to have these tag/subject data in a different field in the vernacular data. 

**Question:**
Could this possibly be an issue with the fetcher? Would it be possible to grab `tags`, and if so, what type of effort would this take?

Also worth noting: YouTube does not seem to display tags in their UI anymore - this info is not visible on the video page. So it could also just be that these tags are gone?

barbarahui:


metadata_fetcher/fetchers/youtube_fetcher.py
```diff
@@ -50,12 +80,6 @@ def check_page(self, http_resp) -> int:
 
         return len(items)
 
-    def increment(self, http_resp):
-        self.write_page = self.write_page + 1
-
-        data = http_resp.json()
-        self.next_page_token = data.get("nextPageToken")
-
```
amywieliczka, line 58: How does next_page_token get set in cases when we're not fetching a playlist, given that this has been removed? I think this would break for not playlist endpoints. 

```python
def increment(self, http_resp):
    self.write_page = self.write_page + 1
    if not is_playlist(self.url):
        data = http_resp.json()
        self.next_page_token = data.get("nextPageToken")
```

metadata_fetcher/fetchers/youtube_fetcher.py
```diff
@@ -19,17 +21,45 @@ def build_fetch_request(self):
...
+                )
+
+            json_response = response.json()
+
+            self.next_page_token = json_response.get("nextPageToken")
+
+            video_ids = [item.get("contentDetails").get("videoId") for item in json_response.get("items",[])]
+            videos_url = f"https://www.googleapis.com/youtube/v3/videos?id={','.join(video_ids)}"
+        else:
+            videos_url = self.url
```
amywieliczka: Hmmm, I think I see what's happening but I think you've broken the YouTube fetcher for non-playlist instances. I'd suggest:

```python
# this is outside of the class
def is_playlist(url)
    if urlparse(url).path == '/youtube/v3/playlistItems':
        return True
    return False
```

```python
# this is a class method
def iterate_playlist(self):
    # The `playlistItems` API endpoint returns incomplete metadata, so we hit it
    # for a page of video ids, and then hit the `videos` endpoint for full metadata.
    # set self.next_page_token here, since we want the next page from the `playlistItems` API, not the next page from the `videos` endpoint
    playlist_content_url = (
            f"{self.url}"
            f"&key={settings.YOUTUBE_API_KEY}"
            f"&part=contentDetails"
            f"&maxResults=50"
    )
    # ...
    # this part can be exactly the same as lines 25-52 above
    # ...
    video_ids = [item.get("contentDetails").get("videoId") for item in json_response.get("items",[])]
    videos_url = f"https://www.googleapis.com/youtube/v3/videos?id={','.join(video_ids)}"
    return videos_url
```

Then, your if/else looks like:
```python
# this is inside the build_fetch_request method
if is_playlist(self.url):
    videos_url = self.iterate_playlist()
else:
    videos_url = self.url.rstrip('/')
```

Then you can also use `is_playlist` elsewhere, too.

metadata_fetcher/fetchers/youtube_fetcher.py
```diff
@@ -19,17 +21,45 @@ def build_fetch_request(self):
...
 
         url = (
-            f"{self.url.rstrip('/')}"
+            f"{videos_url}"
             f"&key={settings.YOUTUBE_API_KEY}"
             f"&part=snippet"
             f"&maxResults=50"
         )
 
-        if self.next_page_token:
```
amywieliczka, line 30: Removing this would break things for not-playlist endpoints

```python
if not is_playlist(self.url) and self.next_page_token:
    url += f"&pageToken={self.next_page_token}"
```

amywieliczka (changes requested):

metadata_fetcher/fetchers/youtube_fetcher.py
barbarahui: In the case of non-playlists, `nextPageToken` doesn't exist in the vernacular, next_page_token gets set to None, and thus the fetching stops. Distinguishing between playlists and non-playlists like this isn't necessary.

metadata_fetcher/fetchers/youtube_fetcher.py
barbarahui: The fetcher actually still works for non-playlist instances. We fetch the first page, see that self.next_page_token is None (this was passed in as a parameter to the class and is never modified), and stop fetching.

metadata_fetcher/fetchers/youtube_fetcher.py
barbarahui: See above: In the case of non-playlists, nextPageToken doesn't exist in the vernacular, next_page_token gets set to None, and thus the fetching stops. Distinguishing between playlists and non-playlists like this isn't necessary.

metadata_fetcher/fetchers/youtube_fetcher.py
amywieliczka: Hmmmm, the `/videos` YouTube api does return a `nextPageToken`: https://developers.google.com/youtube/v3/docs/videos/list 

In our case, the only instance of it is this one-object collection: https://registry.cdlib.org/27022/point-of-pride-the-peoples-view-of-bayviewhunters/ with this API request: https://www.googleapis.com/youtube/v3/videos?id=GCqS7DhJrzA - it only includes one video ID, but a contributor could, theoretically, have a harvest api endpoint including many ids, more than might be returned in a single page. Since the endpoint does return a nextPageToken, it seems reasonable not to delete code that handles it? 

Maybe I'm missing something here.

barbarahui: Ah ok sure, it's true that contributors could have a harvest API endpoint that enumerates > 50 ids. In practice, however, they are providing only one id in the one case where they are using the `videos` endpoint. If they want us to harvest more than one video, they are using the `playlistItems` endpoint.

But OK, I'll take your suggestion to make the fetcher functional for the theoretical use case.

barbarahui:
@amywieliczka This is ready for your re-review

amywieliczka (approved):

metadata_fetcher/fetchers/youtube_fetcher.py
```diff
@@ -19,15 +26,19 @@ def build_fetch_request(self):
         - https://www.googleapis.com/youtube/v3/playlistItems?playlistId={id}
         - https://www.googleapis.com/youtube/v3/videos?id={id}
         """
+        if is_playlist(self.url):
+            videos_url = self.iterate_playlist()
```
amywieliczka: I'm still kinda confused why we are iterating the playlist in `build_fetch_request`, rather than in `increment`, but this makes more sense to me than the last PR.

amywieliczka (dismissed):

metadata_fetcher/fetchers/youtube_fetcher.py
barbarahui: I renamed `iterate_playlist` to `get_videos_url_for_playlist_page`, added comments, and hopefully made things a bit more explicit so that they make more sense? If you're confused, then I'm sure it'll be confusing for whoever needs to modify this down the line!

Does this make more sense now? Am I missing a simpler way to do things?

amywieliczka (approved):

PR #1341, opened 2026-07-21
Validator fixes

### LINKED ISSUE #1342: Rikolti validator bug fixes

barbarahui:
While working on adding a new mapper, I noticed a couple of bugs in the validator code, which are fixed in this PR:

1. The scroll implementation wasn't working, which meant that only the first 1000 indexed record ids were fetched
2. When fetching "initial" indexed data, we were not limiting the query by collection_id. Since we were only querying on calisphere-id, we were fetching indexed records from other collection(s) as the basis of comparison.


amywieliczka:
Isn't calisphere-id the one used in the url? Wouldn't we want to know if we're overwriting another calisphere-id or if there are duplicates?

barbarahui:
No, `id` is the one used in the url

